PR #73, opened 2019-04-17
docs: update name of notification group

MarshallOfSound:
As per Sec WG


MarshallOfSound:
@nornagon I removed it, I think @felixrieseberg is following up somewhere with the new info / reqs / policy

PR #94, opened 2019-05-20
docs: new backport rule

sofianguy:
Adding new backport rule from output of [Releases WG 5/15 meeting](https://github.com/electron/governance/blob/master/wg-releases/meeting-notes/2019-05-15.md#action-items)

### DIFF
--- a/wg-releases/README.md
+++ b/wg-releases/README.md
@@ -40,3 +40,6 @@ See [repos.md](repos.md)
 * **Major Release Cadence Meeting** 1 hour every Thursday at [00:00AM UTC](https://duckduckgo.com/?q=00%3A00+UTC&ia=answer)
 
 Meeting notes may be viewed in [meeting-notes](meeting-notes).
+
+## Feature Backport Requests for Major Versions in Beta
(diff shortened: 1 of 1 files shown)


codebytere:
I think i was absent from this meeting so i may be missing context, but I'm not sure what this rule changes in that I was under the impression that 

> backport requests after Week 3 will only be considered if there is a very good reason.

backport requests _always_ required a good reason even before this?

sofianguy:
@codebytere we are no explicitly saying no backports after Week 3 unless there is a _very_ good reason. Yes all backports, even before Week 3, need to have a good reason. We are a little loose about backports and in the meeting we wanted a concrete rule to point to for future bp requests. 

For example: The latest [backport request](https://github.com/electron/governance/blob/master/wg-releases/meeting-notes/2019-05-15.md#backport-requests)'s reason was addresses user feedback. We voted yes for this one because it was Week 3, but would have voted no if it were requested later.

ckerr:
:+1: on @sofianguy's explanation.

In addition, by writing it down we have something to refer back to / to cite as needed. This could be useful since backports have sometimes been a contentious topic.

PR #124, opened 2019-07-19
docs: enumerate Releases WG removal process

codebytere:
As per discussion on our last meeting, this PR formally adds the process by which sitting members of the Releases WG may be removed. 

cc @electron/wg-releases


sofianguy:
Did we want to add a line item about adding the member back into the working group after a vote?

codebytere:
@sofianguy potentially! what do you think that should look like?

sofianguy:
@codebytere I seem to recall us mentioning in the meeting about after 1 month of inactivity to vote before membership removal. 

Now that I read your current wording, I think that's what you meant. Nevermind!

PR #147, opened 2019-09-12
feat: Docs & Tools WG chair rotation, 2019-09

malept:
We've rotated the chair in the Slack channel, per our rules.


Charles90900:
.github/CODEOWNERS

PR #167, opened 2019-10-03
feat: explain how new participants can join groups

ckerr:
Our project had an influx of new maintainers after the NYC summit :tada: This is the first time we've had an influx since governance was introduced, and there was room for improvement:

1. Some new volunteers were unsure how to join a Working Group.
1. Some [maintainers](https://github.com/electron/governance/#definitions) were unsure about any process around bringing new volunteers into a Working Group.
1. Most Working Groups don't outline requirements for group membership.
1. C&S's role in overseeing maintainer membership caused some delay in people joining.

This PR tries to address these by:

1. Showing the clear steps a new volunteer can follow to join Slack and/or governance
1. Setting a baseline for participation expectations
1. Placing the C&S notification responsibility on the applicant, so that it is part of the process and can happen in parallel with the other participation requirements, reducing any delay

### DIFF
--- a/charter/README.md
+++ b/charter/README.md
@@ -42,9 +42,38 @@ These are our core values, as voted by the maintainers, in order of importance. 
 ## Participation
 
  * [Code of Conduct](../CODE_OF_CONDUCT.md)
+ * [Joining](#joining_governance)
  * [Using Slack](https://github.com/electron/governance/blob/master/policy/slack.md)
  * [Triaging Issues](https://github.com/electron/governance/blob/master/playbooks/README.md)
 <!-- * [Using Pull Requests](FIXME: link to PR etiquette doc) -->
+
+### Joining
+
+Thank you for wanting to contribute to Electron!
+
+If you're not already on the Electron Slack, please sign this
+[Code of Conduct form](https://electronjs.org/maintainers/join).
+The [Community and Safety Working Group](../wg-community-safety) will
+process your application and, if everything is alright, you'll be given
+guest access to Electron's Slack.
+
+To join a [Working Group](README.md#working-groups), you should contact
+an existing member of the group on Electron's Slack about how to join.
+Requirements differ from group to group &ndash; for example,
+[Security](../wg-security) has stricter [rules](../wg-security/membership-and-notifications.md#membership)
+by necessity &ndash; but you're generally expected to show participation
+becoming a full member:
+
+ * Actively contribute to the Working Group's work
+ * Attend 3 out of 6 consecutive meetings
+ * Notify `@community-safety-wg` in Slack about your intent to join _if_
+   you're not already in a Working Group. One benefit of joining a Group
+   is becoming a [maintainer](README.md#definitions), and maintainer
+   membership is handled by C&S.
+
+Once you've done all these things, notify the Working Group and it will decide
+on your membership. This generally involves a vote by existing members, but each
+group has its own process.
 
 ## Working Groups
 


charter/README.md
```diff
@@ -42,10 +42,39 @@ These are our core values, as voted by the maintainers, in order of importance.
...
+To join a [Working Group](README.md#working-groups), you should contact
+an existing member of the group on Electron's Slack about how to join.
+Requirements differ from group to group &ndash; for example,
+[Security](../wg-security) has stricter [rules](../wg-security/membership-and-notifications.md#membership)
+by necessity &ndash; but you're generally expected to show participation
+becoming a full member:
+
+ * Actively contribute to the Working Group's work
+ * Attend 3 out of 6 consecutive meetings
+ * Notify `@community-safety-wg` in Slack about your intent to join _if_
```
felixrieseberg: I want to call out that this is a big shift in the governance of Electron and one I'm not particularly happy with. There's no doubt in the current role of C&S – if there's a complaint on the table, C&S has the power to eject someone from the project.

With this change, we're saying that `C&S` _approval_ is required before a group can welcome a new member. Can someone explain to me why the individual working groups cannot be trusted with electing members?

ckerr: I don't think this is a shift? To my memory, this is consistent with what we discussed in London.

If a person joins a working group, they also become a maintainer and have all the access and permissions associated with that on GitHub and Slack. From the beginning, it was the intent that C&S would administer that. I specifically remember a discussion in London of case of a person who had been ejected from the Electron group for bad behavior in the past, and the need for some oversight to prevent them from getting maintainer permissions just by joining a random WG.

So if the question is "C&S fundamentally shouldn't have a say in who can be a maintainer", I think the next question is "how do we prevent bad actors from abusing the system?" in terms of who has that authority in governance, and the end result may look a lot like C&S.

OTOH if the question is "how can we make it as painless as possible for new contributors to join" ... on that, I think you'll find we're all in agreement. If you have ideas on how to make C&S's oversight more frictionless, I'd love to hear them.

groundwater: It may be worth clarifying, I think this is supposed to _only_ apply to folks new to governance. The C&S ask would be that those individuals _have_ some demonstrated activity before they gain full access to slack and extra access to github. This would put us more in line with other projects. As it stands, someone with zero contributions could _ask to join_ and soon after have access on-par with many long-term contributors.

If someone is already in governance, C&S is not involved.

felixrieseberg: The fact that we're introducing a new section in the charter to document that _is_ a meaningful change and I'm reluctant to accept it as an expression of what's already happening. C&S might be administrating change in permissions, but hearing that C&S also has final say on who gets said permissions is very new to me.

My concern is that we're saying that the individual working groups cannot be trusted with electing members – and that we need a different body to ensure that those working groups cannot willy-nilly introduce new maintainers without oversight. Not just reactive oversight (something bad happened and we need to react), but proactive oversight where the working groups need to ask for permission.

This is just the first wrinkle in this plan, but "3 out of 6" working group meetings could be a _lot of time_ for `wg-security`. We can discuss all day long what the actual requirements should be, but I personally think that we should trust our existing maintainers to be smart and thoughtful when electing new members. We trust them with work and decisions that go far beyond that.

felixrieseberg:
> Most Working Groups don't outline requirements for group membership.

I'd carefully and in the kindest way possible also point out that `C&S` is the only working group that doesn't have membership requirements outlined* 😉 

* `wg-website` doesn't either, but is currently being dismantled.

charter/README.md
groundwater: Joining governance has always been subject to C&S approval.

> we need a different body to ensure that those working groups cannot willy-nilly introduce new maintainers without oversight. Not just reactive oversight (something bad happened and we need to react), but proactive oversight where the working groups need to ask for permission

That has always been the case since our electronhq slack existed, even before governance. Folks were required to submit our _join_ form before gaining access to slack. Entrusting C&S was the formalization of this process.

> we should trust our existing maintainers to be smart and thoughtful when electing new members

We do trust that folks are thoughtful when electing new members. We also trust that folks are thoughtful when creating PRs, but those PRs are still reviewed before merging.

> "3 out of 6" working group meetings could be a lot of time for wg-security

It is my understanding that this is more _feature_ than bug. Our current process allows any WG to admit folks via multi-channel gusts. Aside from disliking the process, perhaps we could document areas where this is preventing groups from getting their work done.

felixrieseberg: > That has always been the case since our electronhq slack existed, even before governance. Folks were required to submit our join form before gaining access to slack. Entrusting C&S was the formalization of this process.

It was my understanding that we added that requirement so that everyone signs the CoC, not to allow a group of humans to have a final opinionated say on whether or not that person is allowed "in".

With `wg-admin`, we made it explicit that members shouldn't be in other working groups due to prevent a clear super-power group. I believe that this change is going against that spirit. With this change `C&S` will essentially have hiring & firing power over the whole project, concentrating more power to a smaller group of people. 

🙏 That said, I feel sufficiently heard, so I'm resolving the conversation and want to thank you all for your time. I still believe that this is ultimately "bad" for the project, but I also understand that working together requires everyone to sail a course they don't favor every now and then.

There are two things I'd like to ask for:

• I expect over-the-top detailed explanations whenever `C&S` decides to veto a member. Simply stating that you will only accept new maintainers who "actively contribute to the Working Group's work" is awfully imprecise and up to personal interpretation.
• This is the _very last_ working group that hasn't outlined its own membership requirements yet. That's a bad look for a group that wants to control other working group's memberships.

PR #169, opened 2019-10-07
chore: Remove nitsakh from the members list

nitsakh:
Remove nitsakh from upgrades and docs-and-tools working groups


codebytere:
@nitsakh can you please rebase?

PR #175, opened 2019-10-10
docs: draft for API Best Practices

zcbenz:
This is a rough draft for API Best Practices, nothing here is settled, the purpose of this PR is to collect opinions and get maintainers together to work out a document that most of us agree with.

This document does not mean to be a JavaScript style guide that defines coding styles or naming rules, this means to describe best practices for designing public Electron APIs.

One thing I'm still not sure is to prefer whether getter/setters or properties, as we have discussed in https://github.com/electron/electron/issues/12745 but still have no decision. I have just listed all options here.

Feel free to commit to this branch to fix grammar/wording.

### DIFF
--- a/wg-api/best-practices.md
+++ b/wg-api/best-practices.md
@@ -94,7 +94,7 @@ win.webContents.session === win.webContents.session
 win.webContents.debugger === win.webContents.debugger
 ```
 
-If an API of the class returns [non-primitives][no-primitive], and the returned values do not always strictly equal each other, it should _not_ be implemented as property.
+If an API of the class returns [non-primitives][primitive], and the returned values do not always strictly equal each other, it should _not_ be implemented as property.
 
 For example:
 
@@ -107,7 +107,7 @@ win.browserViews.push(new BrowserView())
 
 ### Instance Methods
 
-If an API accepts options, or may accept option in future, it should be implemented as method.
+If an API of the class accepts options, or may accept option in future, it should be implemented as method.
 
 For example:
 
@@ -116,9 +116,11 @@ const win = new BrowserWindow(options)
 win.setBounds(bounds, animate)
 ```
 
-### Method vs Property (Option A)
+### Getters and Setters
 
-Unless for good reason, APIs that set/get values should use methods instead of properties.
+If the APIs of the class are used to set/get values, they should be implemented as methods.
+
+If the type of the value being set/get is [primitive][primitive], it should be considered also adding an property interface for the value.
 
 For example:
 
@@ -126,23 +128,10 @@ For example:
 const win = new BrowserWindow(options)
 win.setTitle('str')
 console.log(win.getTitle())
-```
-
-### Method vs Property (Option B)
-
-Unless for good reason, APIs that set/get values should use properties instead of methods.
-
-For example:
-
-```javascript
-const win = new BrowserWindow(options)
+// Property also works.
 win.title = 'str'
 console.log(win.title)
 ```
-
-### Method vs Property (Option C)
-
-There is no requirement on whether to use method or property. _(This chapter should be removed)_
 
 ## Asynchronous APIs
 
@@ -165,4 +154,4 @@ win.hookWindowMessage('MESSAGE', (args...) => {
 ```
 
 [node-buffer-api]: https://medium.com/@jasnell/node-js-buffer-api-changes-3c21f1048f97
-[no-primitive]: https://developer.mozilla.org/en-US/docs/Glossary/Primitive
+[primitive]: https://developer.mozilla.org/en-US/docs/Glossary/Primitive
--- a/wg-api/best-practices.md
+++ b/wg-api/best-practices.md
@@ -120,7 +120,7 @@ win.setBounds(bounds, animate)
 
 If the APIs of the class are used to set/get values, they should be implemented as methods.
 
-If the type of the value being set/get is [primitive][primitive], it should be considered also adding an property interface for the value.
+If the type of the value being set/get is [primitive][primitive], it should be also added as a property interface for the value.
 
 For example:
 
--- a/wg-api/best-practices.md
+++ b/wg-api/best-practices.md
@@ -1,4 +1,168 @@
 # Public API Best Practices
 
-This is a placeholder for a document being drafted by @zcbenz.
-It will be added in a separate PR.
+This is a draft for best practices when designing public APIs.
+
+## Module
+
+### Module Name
+
+If a module exports only one class, the module's name should be the class's name.
+
+For example:
+
+```javascript
+const { BrowserWindow } = require('electron')
+```
+
+If a module includes methods or more than one class, the module's name should be in `lowerCamelCase`.
+
+For example:
+
+```javascript
+const { nativeImage } = require('electron')
+nativeImage.createEmpty()
+```
+
+## Class
+
+### Creating Instance
+
+There are 2 styles for creating instances of classes:
+
+```javascript
+const { BrowserWindow, nativeImage } = require('electron')
+
+// Constructor
+new BrowserWindow(options)
+
+// Factory method
+nativeImage.createFromPath('/path')
+```
+
+#### Constructor
+
+If the class has inheritance relationship with other public classes, constructor should be used.
+
+For example:
+
+```javascript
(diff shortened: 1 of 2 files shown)


wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+// Constructor
+new BrowserWindow(options)
+
+// Factory method
+nativeImage.createFromPath('/path')
+```
+
+#### Constructor
+
+If the class has inheritance relationship with other public classes, constructor should be used.
```
codebytere: nit: `If the class has an inheritance relationship with other public classes, a constructor should be used for that class.`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+
+```javascript
+if (win instanceof TopLevelWindow) {
+  // ...
+}
+```
+
+#### Factory method
+
+If the class accepts different kinds of parameters with options, we should consider using factory methods.
```
codebytere: For consistency: `If the class accepts different kinds of parameters with options, consider using factory methods.`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+For example:
+
+```javascript
+const { nativeImage } = require('electron')
+nativeImage.createEmpty()
+```
+
+## Class
+
+### Creating Instance
```
codebytere: s/`Creating Instance`/`Creating Instances`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+const { BrowserWindow, nativeImage } = require('electron')
+
+// Constructor
+new BrowserWindow(options)
+
+// Factory method
+nativeImage.createFromPath('/path')
+```
+
+#### Constructor
```
codebytere: s/`Constructor`/`Constructors`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+
+This enables users to detect inheritance relationships between classes.
+
+```javascript
+if (win instanceof TopLevelWindow) {
+  // ...
+}
+```
+
+#### Factory method
```
codebytere: s/`Factory method`/`Factory Methods`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+If a module includes methods or more than one class, the module's name should be in `lowerCamelCase`.
+
+For example:
+
+```javascript
+const { nativeImage } = require('electron')
+nativeImage.createEmpty()
+```
+
+## Class
```
codebytere: s/`Classes`/`Class`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
 # Public API Best Practices
 
-This is a placeholder for a document being drafted by @zcbenz.
-It will be added in a separate PR.
+This is a draft for best practices when designing public APIs.
+
+## Module
```
codebytere: s/`Module`/`Modules`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+
+### Module Name
```
codebytere: s/`Module Name`/`Module Names`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+```javascript
+// Using module methods
+const { nativeImage } = require('electron')
+nativeImage.createFromNamedImage('NSImageNameListViewTemplate', [-1, 0, 1])
+
+// Using static methods
+Buffer.from('7468697320697320612074c3a97374', 'hex')
+```
+
+One reason is ambiguities and usability issues, the `Buffer` API is a good example: [Node.js Buffer API Changes][node-buffer-api].
```
codebytere: Let's say:
```
Factory methods help reduce ambuguity and usability issues. The `Buffer` API provides a good example of this: [Node.js Buffer API Changes][node-buffer-api].
````

File: wg-api/best-practices.md [hunk 10]
```diff
@@ -1,4 +1,168 @@
...
+
+Another reason is, due to lack of function overloading, JavaScript class's constructor can only be implemented with a single C++ function, and it is error-prone to parse different kinds of parameters with C++ manually. While factory methods can be perfectly mapped to C++ functions, making the code much easier to read and maintain.
```
codebytere: How about: 
> Additionally, due to lack of function overloading, JavaScript class constructors can only be implemented with a single C++ function and are error-prone when parsing different kinds of parameters with C++. Conversely, factory methods can be perfectly mapped to C++ functions, thereby making code much easier to read and maintain.

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+
+### Instance Event
```
codebytere: s/`Instance Event`/`Instance Events`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+
+If the class generates events, it should inherit from `EventEmitter` and use its interface for emitting events.
+
+### Instance Property
```
codebytere: s/`Instance Property`/`Instance Properties`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+
+If an API of the class returns _non-assignable_ `Object`, and the returned objects always strictly equal each other, the API should be implemented as property.
```
codebytere: s/`returns _non-assignable_`/`returns a _non-assignable_`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+For example:
+
+```javascript
+// Should NOT do these, as they have no effect.
+const win = new BrowserWindow(options)
+win.bounds.x = 42
+win.browserViews.push(new BrowserView())
+```
+
+### Instance Method
```
codebytere: s/`Instance Method`/`Instance Methods`

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+If an API accepts options, or may accept option in future, it should be implemented as method.
+
+For example:
+
+```javascript
+const win = new BrowserWindow(options)
+win.setBounds(bounds, animate)
+```
+
+### Method vs Property (Option A)
```
codebytere: I'll leave edit to this section until after we've discussed them as a WG :)

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+console.log(win.title)
+```
+
+### Method vs Property (Option C)
+
+There is no requirement on whether to use method or property. _(This chapter should be removed)_
+
+## Asynchronous APIs
+
+If an API returns _one_ result asynchronously for each call, it should use `Promise`.
```
codebytere: Let's say:
>If an API returns _one_ result asynchronously for each call, it should be implemented with `Promises`.

wg-api/best-practices.md
```diff
@@ -1,4 +1,168 @@
...
+
+For example:
+
+```javascript
+const win = new BrowserWindow(options)
+const result = await win.webContents.executeJavaScript('void 0')
+```
+
+For subscription-style API that returns results for multiple times, it should still use callbacks.
```
codebytere: Let's say:
> Subscription-style APIs that returns results multiple times should still be implemented with callbacks.

codebytere:
Thanks for getting this started @zcbenz 🙇‍♀

ckerr:
Maybe we should do something with this PR? :smiley_cat:

zcbenz:
Updated the doc as per last wg-api meeting's discussion.

wg-api/best-practices.md
```diff
@@ -1,4 +1,157 @@
...
+```javascript
+const win = new BrowserWindow(options)
+win.setBounds(bounds, animate)
+```
+
+### Getters and Setters
+
+If the APIs of the class are used to set/get values, they should be implemented as methods.
+
+If the type of the value being set/get is [primitive][primitive], it should be considered also adding an property interface for the value.
```
jkleinsc: Considering we want consistency, I think we should be more prescriptive here.  Maybe something like:
```suggestion
If the type of the value being set/get is [primitive][primitive], it should be also added as a property interface for the value.
```

PR #191, opened 2019-10-24
doc: add spec-proposal process docs

codebytere:
This is an initial spike for a document design process overseen by the API Working Group. None of this has been decided upon and all can change; this is simply to start this work into motion and encourage discussion and feedback.

cc @electron/wg-api

### DIFF
--- a/wg-api/spec-documents/api-spec-template.md
+++ b/wg-api/spec-documents/api-spec-template.md
@@ -1,0 +1,16 @@
+# API Spec Template
+
+## Summary
+<!--- 1-3 sentence description of what is changing and why. --->
+
+## Platforms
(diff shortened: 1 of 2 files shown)


wg-api/spec-documents/api-spec-template.md
```diff
@@ -0,0 +1,16 @@
+# API Spec Template
+
+## Summary
+<!--- 1-3 sentence description of what is changing and why. --->
+
+## Platforms
+<!--- What platforms will be affected? Options include macOS, Windows, Linux, or any combination thereof. --->
+
+## Impetus
```
VerteDinde: Really glad that this is its own section; I think this is important for justification.

PR #197, opened 2019-10-29
docs: add more Chair responsibilities

codebytere:
This PR adds and explicates a few more areas of responsibility for each WG Chair that have been implicit but would be likely better made explicit!

### DIFF
--- a/charter/README.md
+++ b/charter/README.md
@@ -104,13 +104,13 @@ Chairs have responsibility to all participants in transparently reporting the gr
 
 Chairs also have the following regular duties within their Working Groups:
 * Collect agenda items prior to each meeting to ensure a smooth and focused Working Group meeting.
-* Ensure that the Working Group is notified of the agenda for the meeting in a timely manner advance of the meeting, and that if there is no agenda, the meeting is altered or cancelled accordingly.
+* Ensure that the Working Group is notified of the agenda for the meeting in a timely manner advance of the meeting.
 * Attend and lead each Working Group meeting.
     * If the Chair is unable to attend a given meeting, they should delegate a pre-appointed person to lead the meeting in their stead.
 * Ensure meeting notes are uploaded to the `governance` repo in the Working Group's respective sub-folder following each meeting.
 * Check into persistent issues present on a given Working Group's repositories of responsibility (ex. bot failures)
 
-Each Working Group may also establish more specific Chair responsibilities within their domains.
+In addition to these reponsibilities, a Chair may have additional responsibilities specific to their Working Group.
 
 ### Reaching Agreement
 
--- a/charter/README.md
+++ b/charter/README.md
@@ -109,6 +109,7 @@ Chairs also have the following regular duties within their Working Groups:
     * If the Chair is unable to attend a given meeting, they should delegate a pre-appointed person to lead the meeting in their stead.
 * Ensure meeting notes are uploaded to the `governance` repo in the Working Group's respective sub-folder following each meeting.
 * Check into persistent issues present on a given Working Group's repositories of responsibility (ex. bot failures)
+* Ensure that the WG roster list is an accurate reflection of current member activity.
 
 In addition to these reponsibilities, a Chair may have additional responsibilities specific to their Working Group.
 
--- a/charter/README.md
+++ b/charter/README.md
@@ -102,6 +102,16 @@ Chairs have responsibilities to their parent Working Group: they ensure that the
 
 Chairs have responsibility to all participants in transparently reporting the group's activity.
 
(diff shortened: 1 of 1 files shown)


charter/README.md
```diff
@@ -102,6 +102,16 @@ Chairs have responsibilities to their parent Working Group: they ensure that the
...
 Chairs have responsibility to all participants in transparently reporting the group's activity.
 
+Chairs also have the following regular duties within their Working Groups:
+* Collect agenda items prior to each meeting to ensure a smooth and focused Working Group meeting.
+* Notify the Working Group channel at least 12 hours before the meeting, outlining the agenda.
+* Attend and lead each Working Group meeting.
+    * If the Chair is unable to attend a given meeting, they should delegate a pre-appointed person to lead the meeting in their stead.
+* Ensure meeting notes are uploaded to the `governance` repo in the Working Group's respective sub-folder following each meeting.
+* Check into persistent issues present on a given Working Group's repositories of responsibility (ex. bot failures)
+
```
sofianguy: Does it make sense to add a bullet for something about action items from meetings:

> Ensure WG `Action Items` are actively being worked on

charter/README.md
```diff
@@ -102,6 +102,16 @@ Chairs have responsibilities to their parent Working Group: they ensure that the
 
 Chairs have responsibility to all participants in transparently reporting the group's activity.
 
+Chairs also have the following regular duties within their Working Groups:
+* Collect agenda items prior to each meeting to ensure a smooth and focused Working Group meeting.
+* Notify the Working Group channel at least 12 hours before the meeting, outlining the agenda.
```
bnb: Is it worth taking this work off a human's plate and automating it instead?

charter/README.md
codebytere: i think every working group does action items differently, so i would prefer not to be prescriptive about this i think!

charter/README.md
codebytere: i think i might update this, since i think it's more complex than that:
> Ensure that the Working Group is notified of the agenda for the meeting in a timely manner advance of the meeting, and that if there is no agenda, the meeting is altered or cancelled accordingly.

codebytere:
@malept
> Should there be a point about keeping the WG roster up-to-date?

I feel this may be too specific for this README, as it's perhaps more of an imp;ementation detail?

charter/README.md
```diff
@@ -102,6 +102,16 @@ Chairs have responsibilities to their parent Working Group: they ensure that the
...
 
+Chairs also have the following regular duties within their Working Groups:
+* Collect agenda items prior to each meeting to ensure a smooth and focused Working Group meeting.
+* Ensure that the Working Group is notified of the agenda for the meeting in a timely manner advance of the meeting, and that if there is no agenda, the meeting is altered or cancelled accordingly.
+* Attend and lead each Working Group meeting.
+    * If the Chair is unable to attend a given meeting, they should delegate a pre-appointed person to lead the meeting in their stead.
+* Ensure meeting notes are uploaded to the `governance` repo in the Working Group's respective sub-folder following each meeting.
+* Check into persistent issues present on a given Working Group's repositories of responsibility (ex. bot failures)
+
+Each Working Group may also establish more specific Chair responsibilities within their domains.
```
groundwater: I would prefer to keep a single _chair_. The one person ultimately responsible, but that person may delegate most, if not all, of their duties to other members of the group. The intent is to avoid the bystander-effect. Two chairs can never be sure who is ultimately responsible for what. I have no objection to the idea, just the name. Using _sub-chair_, _chief_, _lead_, or anything else would be fine.

charter/README.md
```diff
@@ -102,6 +102,16 @@ Chairs have responsibilities to their parent Working Group: they ensure that the
 
 Chairs have responsibility to all participants in transparently reporting the group's activity.
 
+Chairs also have the following regular duties within their Working Groups:
+* Collect agenda items prior to each meeting to ensure a smooth and focused Working Group meeting.
+* Ensure that the Working Group is notified of the agenda for the meeting in a timely manner advance of the meeting, and that if there is no agenda, the meeting is altered or cancelled accordingly.
```
groundwater: Note that persistent cancelling of meetings should be seen as a sign the working group may need restructuring, reorganization, etc.

malept:
> > Should there be a point about keeping the WG roster up-to-date?
> 
> I feel this may be too specific for this README, as it's perhaps more of an implementation detail?

I don't think it's too specific. It's up to the specific WG to decide how they add / remove members (e.g., #200) but given the privileges involved in being a governance member, it should be explicitly stated somewhere.

akhlak125:
Digital services regard to login

PR #199, opened 2019-10-29
docs: add release process change RFC

codebytere:
Adds the RFC for the release process change we've since implemented, but whose document was never uploaded. For posterity, we should keep these in the WG folder.

cc @jkleinsc (for doing the work to implement this ✨) @electron/wg-releases

### DIFF
--- a/wg-releases/RFCs/release-process-change.md
+++ b/wg-releases/RFCs/release-process-change.md
@@ -1,0 +1,23 @@
+# RFC: Release Process Change
+
+**Author:** Shelley Vohr
+
+## What We Do Now
+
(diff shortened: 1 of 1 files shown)


wg-releases/RFCs/release-process-change.md
```diff
@@ -0,0 +1,23 @@
+# RFC: Release Process Change
+
+**Author:** Shelley Vohr
+
+## What We Do Now
+
+Right now, we create a new branch for every new minor release line. For example, there current exist branches for `4-0-x`, `4-1-x`, and a of recently `4-2-x`.
```
malept:
```suggestion
Right now, we create a new branch for every new minor release line. For example, there current exist branches for `4-0-x`, `4-1-x`, and as of recently, `4-2-x`.
```

PR #204, opened 2019-11-05
docs: spec document: MessagePorts

nornagon:
imported from https://hackmd.io/bwzFRWZzRIyewIBc1Vf7DQ

### DIFF
--- a/wg-api/spec-documents/message-ports.md
+++ b/wg-api/spec-documents/message-ports.md
@@ -142,7 +142,8 @@ function beginWork() {
 ```
 
 *worker_renderer.js*
-```javascript==ipcRenderer.on('register', (e, {port}) => {
+```javascript=
+ipcRenderer.on('register', (e, {port}) => {
   port.onmessage = handleOperation(port)
 })
 
--- a/wg-api/spec-documents/message-ports.md
+++ b/wg-api/spec-documents/message-ports.md
@@ -1,0 +1,206 @@
+# MessagePorts in Electron
+
+## Problem
+
+It's common for Electron apps to use a hidden renderer for a variety of reasons: to execute untrusted code in a sandboxed background window, to relegate CPU-intensive work to another process, or to isolate crash-prone components. Currently, it's non-trivial to communicate directly from one renderer process to another unrelated process (i.e. one not opened by `window.open`): either one must relay via the main process through `sendToFrame`, or use a 3rd-party native module such as `node-ipc`.
+
+I propose that we reuse and extend the already-existing web-standard [Channel Messaging API][channel-messaging-api] to facilitate communication between renderers, and as a bonus, take the opportunity to add a powerful new IPC primitive to Electron: message channels.
+
+## Background
+
+### The Channel Messaging API
+
+The [Channel Messaging API][channel-messaging-api] is a web standard that builds on `postMessage` by allowing the creation of multiple independent channels across which messages can be sent. On the web, it can be used to communicate with Web Workers, Service Workers, or between windows (or even within the same window, somewhat redundantly).
+
+Here's a short example of usage on the Web:
+
+```javascript=
+const {port1, port2} = new MessageChannel();
+port1.onmessage = (e) => console.log('port1 got message', e.data);
+someOtherWindow.postMessage(
+    `here's the other end of the channel`, '*', [port2]);
+
+// In the other window
+window.onmessage = (e) => {
+  const port2 = e.ports[0]
+  port2.postMessage('send a message over the channel')
+  // port1.onmessage will be triggered
+}
+```
+
+Under the hood, this is a light wrapper around Mojo pipes and the same algorithms that `window.postMessage` uses.
+
+### Node.js worker threads
+
+Node.js took inspiration from the Channel Messaging API when they designed the [communication system for worker threads](https://nodejs.org/dist/latest-v12.x/docs/api/worker_threads.html#worker_threads_class_messagechannel).
+
+### Transferables
+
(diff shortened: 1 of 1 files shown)


wg-api/spec-documents/message-ports.md
```diff
@@ -0,0 +1,206 @@
...
+
+* `channel` String
+* `message` any
+* `options` Object (optional)
+  * `transfer` any[] - a list of transferable objects to be sent with the message.
+
+Similar to `WebContents.send`, but allows transfering transferable objects.
+
+> [name=Samuel Attard]
+> I'm trying to figure out if there's a way for this to be non-breaking integrated into `.send` to avoid the confusion around two different ways to send IPC.  Can't see an easy way out there though.
```
nornagon: Yeah neither. The best I came up with is that any MessagePorts sent through `send` get implicitly transferred. It's not great though.

nornagon: @MarshallOfSound

wg-api/spec-documents/message-ports.md
```diff
@@ -0,0 +1,206 @@
...
+###### Event: 'close'
+
+Emitted when the channel is closed.
+
+::::warning
+:warning: This event is not part of the Web Messaging API.
+::::
+
+> [name=Samuel Attard]
+> Is there a way to tell if a message port has been closed _after_ the fact.  E.g. `port.isClosed` or something?  Otherwise I assume `postMessage` will throw if the port is closed and you try to send a message.
```
nornagon: nope. none of the above, i believe.

wg-api/spec-documents/message-ports.md
```diff
@@ -0,0 +1,206 @@
...
+function beginWork() {
+  workerConnection.postMessage({operation: 'foo', data: [...]})
+  workerConnection.onmessage = (message) => {
+    // ... do something with the message from the worker renderer ...
+  }
+}
+```
+
+*worker_renderer.js*
+```javascript==ipcRenderer.on('register', (e, {port}) => {
```
jkleinsc:
```suggestion
```javascript=
ipcRenderer.on('register', (e, {port}) => {
```

wg-api/spec-documents/message-ports.md
```diff
@@ -0,0 +1,206 @@
...
+
+worker.loadURL(
+  'data:text/html,<script src="worker_renderer.js"></script>')
+  
+ui.loadURL(
+  'data:text/html,<script src="ui_renderer.js"></script>')
+  
+const {port1, port2} = new MessageChannel
+
+worker.webContents.postMessage('register', {port: port1}, [port1])
```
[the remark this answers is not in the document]
nornagon: Yeah, this is just how `postMessage` works. Actually, when using the real `postMessage`, the event on the other end contains an `event.ports` property so you can fetch the transferred ports even without referencing them in the message object, which I suppose we could add.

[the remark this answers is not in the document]
jkleinsc: @nornagon I like the idea of mimicking `postMessage` as much as we can.

wg-api/spec-documents/message-ports.md
nornagon: I propose we patch in an `onclose` event. It's pretty straightforward code-wise: https://github.com/electron/electron/pull/22532 and has some indicators of support from browsers: https://github.com/w3ctag/design-reviews/issues/269#issuecomment-407584290

wg-api/spec-documents/message-ports.md
```diff
@@ -0,0 +1,207 @@
...
+When this message is received in the main process, any `MessagePort` objects transferred will be deserialized as `ipcMain.MessagePort`.
+
+#### Class: WebContents
+
+##### `WebContents.postMessage(channel, message[, options])`
+
+* `channel` String
+* `message` any
+* `options` Object (optional)
+  * `transfer` any[] - a list of transferable objects to be sent with the message.
```
itsananderson: There's a mismatch between this spec and the current PR https://github.com/electron/electron/pull/22532 which just supports `[list, of, transferables]`. I like the spec version better for a couple reasons:

1) More robust against adding more options in the future
2) More self documenting to explicitly say `{ transfer: [a, b, c] }`. I think this will especially help reduce confusion when we support passing ports in the `message`.

Looking at the history of the [web spec](https://html.spec.whatwg.org/multipage/web-messaging.html#posting-messages), it looks like the `postMessage(message, [transferables])` API was added first, and then the `postMessage(message, { transfer })` API was added later (they introduced a `targetOrigin` property that probably made them realize positional arguments can be painful for back-compat).

wg-api/spec-documents/message-ports.md
nornagon: MessagePorts don't begin receiving events (including the "closed" event) until `.start()` is called on them. There's no race condition here.

wg-api/spec-documents/message-ports.md
nornagon: I'm also keen to support the object as well as the array. It's listed in the PR as something to be done in a followup change.

wg-api/spec-documents/message-ports.md
```diff
@@ -0,0 +1,207 @@
...
+worker.loadURL(
+  'data:text/html,<script src="worker_renderer.js"></script>')
+  
+ui.loadURL(
+  'data:text/html,<script src="ui_renderer.js"></script>')
+  
+const {port1, port2} = new MessageChannel
+
+worker.webContents.postMessage('register', {port: port1}, [port1])
+ui.webContents.postMessage('worker_connection', {port: port2}, [port2])
```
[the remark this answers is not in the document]
nornagon: Correct, `start()` is what allows preventing this race condition.

Additionally, messages are queued until `start` is called, so you can do something like:

```js
const {port1, port2} = new MessageChannel
port1.postMessage('hi there')
ipcRenderer.postMessage('port', null, [port2])
```

and the `'hi there'` message will be queued until `port2.start()` is called in the main process (or wherever it ends up).

PR #208, opened 2019-11-05
docs: spec for native window.open changes

loc:
Originally here: https://hackmd.io/6-Bm5_7VTZKQobqQn0M8kg?view

I've culled down the original to only include Option A, changed the event names, and added some additional context. The functionality is the same as originally written.

### DIFF
--- a/wg-api/spec-documents/better-native-window-open.md
+++ b/wg-api/spec-documents/better-native-window-open.md
@@ -25,76 +25,65 @@ As we align more with Chrome's process model, `BrowserWindowProxy` will become l
 
 All platforms
 
-## Proposal
+## API Design
 
-### Summary
+**Instance method `webContents.setWindowOpenOverride(handler)`**
 
-At a high level, I propose we add an event, `will-create-child-web-contents`, so that we can cancel the renderer-initialized `WebContents` creation earlier than we do now. Additionally, for improved clarity and ergonomics, I suggest we change the name of the `WebContents`'s  `new-window` event to `will-create-child-window` and slightly change its mechanics. 
+- `handler` Function
+    - `windowOpenInfo` Object
+        - `url` String
+        - `frameName` String
+        - `disposition` String
+        - `parsedOptions` BrowserWindowConstructorOptions
+        - `additionalFeatures` String[]
 
-I would guess that most classes of app would need to subscribe to zero or one of these events. Only the most complex would need both (Slack! Teams!).
+    Returns `Boolean | BrowserWindowConstructorOptions`
 
-### API Design
+Allows consumers to control whether a window is created and, if it should be created, with which `BrowserWindow` options. Unlike the current implementation, it allows for setting `webPreferences` on child windows.
 
-#### `will-create-child-web-contents`
+Returning `true` will result in the default behavior, a native window will be created and returned to the caller in the renderer. Returning `false` will prevent creation of the window and return `null` to the caller. Returning `BrowserWindowConstructorOptions` will result in the default behavior, but with the provided options used when initializing the `BrowserWindow`.
 
-Addresses parts of **use case #1** wherein the consumer wants to stop a renderer's `window.open` call from succeeding.
-
-**Note:** this will not fire for an iframe, a potential point of confusion.
-
-**Possible implementation**
-
-There are already delegate methods that we could hook into to emit this event before the `WebContents` is created, no patching necessary: `ShouldCreateWebContents` (7-0-x) and `IsWebContentsCreationOverridden` (master). 
-
-**Example**
-
-```javascript
-mainWindow.webContents.on(
-  'will-create-child-web-contents',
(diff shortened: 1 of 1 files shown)


wg-api/spec-documents/better-native-window-open.md
```diff
@@ -0,0 +1,105 @@
...
+**Possible implementation**
+
+There are already delegate methods that we could hook into to emit this event before the `WebContents` is created, no patching necessary: `ShouldCreateWebContents` (7-0-x) and `IsWebContentsCreationOverridden` (master). 
+
+**Example**
+
+```javascript
+mainWindow.webContents.on(
+  'will-create-child-web-contents',
+  (event, url, frameName, disposition, options, additionalFeatures) =>
```
nornagon: It seems like it would be useful to know what WebContents the new WebContents will child of.

wg-api/spec-documents/better-native-window-open.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Platforms
+
+All platforms
+
+## Proposal
+
+### Summary
+
+At a high level, I propose we add an event, `will-create-child-web-contents`, so that we can cancel the renderer-initialized `WebContents` creation earlier than we do now. Additionally, for improved clarity and ergonomics, I suggest we change the name of the `WebContents`'s  `new-window` event to `will-create-child-window` and slightly change its mechanics.
```
nornagon: what's the difference between `will-create-child-web-contents` and `will-create-child-window`? does the latter only get called for `target=_blank` whereas the former will also be created for <webview> or <iframe> or other kinds of navigations?

wg-api/spec-documents/better-native-window-open.md
loc: I believe it will always be `mainWindow.webContents`, i.e. the one you're listening to?

nornagon: ah, okay then!

loc:
With helpful input from Sam, I've revised the proposal to something that we think is easier to implement and use. Take another look!

wg-api/spec-documents/better-native-window-open.md
```diff
@@ -0,0 +1,94 @@
...
+
+- `handler` Function
+    - `windowOpenInfo` Object
+        - `url` String
+        - `frameName` String
+        - `disposition` String
+        - `parsedOptions` BrowserWindowConstructorOptions
+        - `additionalFeatures` String[]
+
+    Returns `Boolean | BrowserWindowConstructorOptions`
```
nornagon: Hmm, I don't love returning the constructor options here, that's kind of a weird API. It seems more natural that if you override this function, you're responsible for creating the new window yourself. i.e., the default behavior will be to cancel the new window, and you have to do something to make that not the case. i.e. if you want to open a window as if you hadn't overridden this, you'd do:

```
webContents.setWindowOpenOverride((info) => {
  new BrowserWindow(info.parsedOptions)
})
```

And if we do things that way, it seems sort of odd not to have this be a regular event with an event handler that you'd call `event.preventDefault()` on, no?

wg-api/spec-documents/better-native-window-open.md
```diff
@@ -0,0 +1,94 @@
...
+
+Allows consumers to control whether a window is created and, if it should be created, with which `BrowserWindow` options. Unlike the current implementation, it allows for setting `webPreferences` on child windows.
+
+Returning `true` will result in the default behavior, a native window will be created and returned to the caller in the renderer. Returning `false` will prevent creation of the window and return `null` to the caller. Returning `BrowserWindowConstructorOptions` will result in the default behavior, but with the provided options used when initializing the `BrowserWindow`.
```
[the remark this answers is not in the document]
MarshallOfSound: (a) invalid return type (b) will be cast to false I'm pretty sure.

wg-api/spec-documents/better-native-window-open.md
MarshallOfSound: We can't construct a new window at this point as the webcontents hasn't been made yet.  The window can't be made until chromium has handed us the webcontents.

wg-api/spec-documents/better-native-window-open.md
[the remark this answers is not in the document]
nornagon: (a) yes, but most Electron apps are JS not TS :)

(b) let's document that then.

wg-api/spec-documents/better-native-window-open.md
loc: It is weird, but it's the least weird we could come up with, given the constraints that Sam mentioned. The API you describe is basically the "ideal" version I laid out in the original writeup. We came to the reasonable consensus that it was too hard for the moment.

loc:
Closing for now, pending some updates to this doc.

Chynia828:
No me?

PR #227, opened 2020-01-22
chore: remove @brenca from upgrades wg member list

nornagon:


codebytere:
Note that we should also remove him and regenerate the upgrade rotations

nornagon:
@codebytere i believe the upgrade rotations were already regenerated to remove him :)

PR #229, opened 2020-01-29
docs: clipboard.getNativeFormat() and clipboard.writeBuffers()

coreh:
Hey folks 👋 

Here's an initial version of the specs for the proposed `clipboard.getNativeFormat()` and `clipboard.writeBuffers()` API methods that we started working on during the contributor training

cc @itsananderson @loc

### DIFF
--- a/wg-api/spec-documents/clipboard-getnativeformat.md
+++ b/wg-api/spec-documents/clipboard-getnativeformat.md
@@ -1,0 +1,35 @@
+# `clipboard.getNativeFormat(portableFormat)`
+
+## Summary
+A new `clipboard.getNativeFormat(portableFormat)` API method is introduced. It returns a `String` representing the underlying _native_ clipboard type for the supplied _portable_ format.
+
+## Platforms
+macOS, Windows, Linux
+
+## Impetus
+
+Each operating system supported by Electron utilizes a different convention to identify the data format of the clipboard contents: macOS uses [UTIs](https://en.wikipedia.org/wiki/Uniform_Type_Identifier), Linux uses [MIME types](https://en.wikipedia.org/wiki/MIME) and Windows uses [numeric values](https://docs.microsoft.com/en-us/windows/win32/dataxchg/standard-clipboard-formats).
+
+To increase portability, Electron's `clipboard` API exposes read/write access to several high-level _portable_ clipboard formats (via `write()`, `writeText()`, `readImage()` etc). Lower-level access to the _native_ formats is also provided via `Buffer`-based read/write APIs (via `readBuffer()`, `writeBuffer()`)
+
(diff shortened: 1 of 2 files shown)


wg-api/spec-documents/clipboard-getnativeformat.md
```diff
@@ -0,0 +1,35 @@
...
+## API Design
+
+The new API is comprised of a single method:
+
+```markdown
+### `clipboard.getNativeFormat(portableFormat)`
+
+* `portableFormat` String - One of the supported native formats: `text`, `bookmark`, `html`, `rtf`, `image`
+
+Returns `String` - An opaque, platform-dependent value representing a clipboard format (e.g. `text/plain`, `public.text` or `13`)
```
nornagon: Presumably the only useful application of this value would be to pass it into the new `clipboard.writeBuffers` (or existing `clipboard.writeBuffer`) function, yes?

So this would be used like:
```js
clipboard.writeBuffers({
  [clipboard.getNativeFormat('text')]: Buffer.from(...),
  [myCustomFormat[os.platform()]]: Buffer.from(...),
})
```

That seems a little awkward to me, though I'm not sure I have a better suggestion. Here are some random thoughts, maybe useful, maybe not.

1. Given that custom formats will also necessarily be platform-specific, how do we expect users of the API to handle those? Should we have some examples to guide best practices?
2. These values are effectively constants. Should we expose them as such? e.g. `clipboard.format.TEXT`
3. Could we use JS Symbols effectively here...? e.g. if `clipboard.format.TEXT` was a special symbol, that could distinguish it from custom formats while avoiding the collision problem.

itsananderson:
I wrote up a really basic Fiddle for testing the Chromium feature, in case it's a helpful jumpstart for anyone: https://gist.github.com/2802c6a3744a31d3a81961bc118632d2 - I didn't do anything fancy to write custom binary formats, but it should be simple to add.

Exposing a compatible API in the main process does seem useful, especially since renderer processes are restricted to reading/writing the clipboard when they have user focus. We may need to make small adjustments to the API, like swapping Buffers for Blobs for binary clipboard values.

henrit:
Chromium's raw clipboard API was [abandoned last June](https://bugs.chromium.org/p/chromium/issues/detail?id=897289#c65). Its implementation has been [scrubbed](https://bugs.chromium.org/p/chromium/issues/detail?id=1217643) from the chromium codebase.

There is now instead a proposal for standardizing [Pickling for Async Clipboard API](https://chromestatus.com/feature/5649558757441536), but it is explicitly designed to prevent setting arbitrary formats on the clipboard in the ways a native app would.

This leaves Electron users in the position of continuing to put together one-off node.js native clipboard modules, or waiting for this proposed change to land.

PR #232, opened 2020-02-06
docs: add print API consistency proposal

codebytere:
This PR is a proposal for the following:

`webContents.print()` and `webContents.printToPDF()` should pass identical parameter types in the same shape. Parameters which differ between the two will be changed to remove differences and prevent unwanted bugs. `webContents.print()` will be Promisified.

cc @electron/wg-api

### DIFF
--- a/wg-api/spec-documents/wc-print-sync.md
+++ b/wg-api/spec-documents/wc-print-sync.md
@@ -1,0 +1,46 @@
+# Print Functionality Synchronization
+
+## Summary
+
+`webContents.print()` and `webContents.printToPDF()` should pass identical parameter types in the same shape. Parameters which differ between the two will be changed to remove differences and prevent unwanted bugs. `webContents.print()` will be Promisified.
+
+## Platforms
+
+All.
+
+## Impetus
+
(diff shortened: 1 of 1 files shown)


wg-api/spec-documents/wc-print-sync.md
```diff
@@ -0,0 +1,46 @@
...
+
+`webContents.print()` and `webContents.printToPDF()` should pass identical parameter types in the same shape. Parameters which differ between the two will be changed to remove differences and prevent unwanted bugs. `webContents.print()` will be Promisified.
+
+## Platforms
+
+All.
+
+## Impetus
+
+`webContents.print()` and `webContents.printToPDF()` pass nearly identical sets of parameters, but given that they were implemented at different times, there is a small delta in the shape in which those parameters are passed in. Our argument sanitization for `webContents.print()` lived natively, and for `webContents.printToPDF()` lives in JavaScript. I plan to refactor these two functions to move all sanitization into JavaScript for the following primary reasons:
```
felixrieseberg: > two functions to move all sanitization into JavaScript for the following primary reasons

Those functions should never be called in high frequency, making this move super sensible 👍

codebytere:
@nornagon updated - should be good to go :)

PR #235, opened 2020-02-12
docs: fix meeting time for Ecosystem WG

erickzhao:
This might have been a Daylight Savings thing, but it seems like the UTC time of our meetings is off. We meet at 9:00 AM PT, which is 5:00 PM UTC.


malept:
This was definitely a DST thing... we're going to have to either change this in a few months, or we figure out a better way to represent the time.

PR #254, opened 2020-03-19
docs: views API: support setting content view

zcbenz:
This spec proposes to make parts of `views` related APIs public, which should be enough to implement a simple splash screen for apps.

```js
const {app, nativeImage, webContents, ImageView, TopLevelWindow, WebContentsView} = require('electron')

app.once('ready', () => {
  const win = new TopLevelWindow({})
  const imageView = new ImageView()
  imageView.setImage(__dirname + '/splash.png')
  win.setContentView(imageView)

  const contents = webContents.create({})
  const web = new WebContentsView(contents)
  contents.once('did-finish-load', () => win.setContentView(web))
  contents.loadURL('https://github.com/')
})
```

### DIFF
--- a/wg-api/spec-documents/views-api-1-content-view.md
+++ b/wg-api/spec-documents/views-api-1-content-view.md
@@ -1,0 +1,100 @@
+# Views API (Part 1): Support Setting Content View
+
+## Summary
+
+This spec proposes to make parts of `views` related APIs public, which should be
+enough to implement a simple splash screen for apps.
+
+(`views` is the name of Chromium's UI library.)
+
+## Platforms
+
+All.
+
+## Impetus
+
+I would like to extend Electron's APIs to provide ability to build native UI,
+but it would too much work to do it all at once. So I plan to do it piece by
+piece, and with each addition the apps should be able to do something that was
+impossible before.
+
+With this spec, users would be able to change the content view of windows. One
+possible use is to show a splash image when the app is loading, and change the
+splash image to WebContents after loading is finished.
+
+```js
+const {app, nativeImage, webContents, ImageView, TopLevelWindow, WebContentsView} = require('electron')
+
+app.once('ready', () => {
+  const win = new TopLevelWindow({})
+  const imageView = new ImageView()
+  imageView.setImage(__dirname + '/splash.png')
+  win.setContentView(imageView)
+
+  const contents = webContents.create({})
+  const web = new WebContentsView(contents)
+  contents.once('did-finish-load', () => win.setContentView(web))
+  contents.loadURL('https://github.com/')
+})
+
+app.on('window-all-closed', () => {
+  app.quit()
+})
+```
+
+## API Design
+
+There are several new APIs in this spec, most of them are actually internal
+APIs that already being used. The main work would be documenting them and making
+them public.
+
+### `TopLevelWindow`
+
+The `TopLevelWindow` class represents a native window, which implements all
+window related APIs and has an empty content view.
+
+The `BrowserWindow` class inherits from `TopLevelWindow` and uses
+`WebContentsView` as its content view.
+
+### `TopLevelWindow.setContentView(view)`
+
+This changes the content view of the window.
+
+The `view` is an instance of `View` class, with this spec there will be 2 types
+of `View`: `WebContentsView` and `ImageView`, and they inherit from the `View`
+class.
+
+### `View`
+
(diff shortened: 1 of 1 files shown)


itsananderson:
This seems like a super useful set of APIs to unlock. Very exciting!

If I want to use `TopLevelWindow` rather than `BrowserWindow`, how much glue code am I going to need to implement myself? One example that occurs to me is that a `BrowserWindow` automatically updates its window title when the title of the `WebContents` changes. Would we also expect that to happen automatically if you embed a `WebContentsView` in a `TopLevelWindow`, or would you need to listen for title changes in the `WebContents` and then propagate to the `TopLevelWindow`?

I don't have a strong opinion about how much of that should happen "for free", so long as the delta between `BrowserWindow` and `TopLevelWindow` is understood. We'll want to documented the differences and provide `TopLevelWindow` APIs that allow people to switch without regressing the behavior they had with `BrowserWindow`

zcbenz:
@itsananderson The `BrowserWindow` inherits from `TopLevelWindow` so you don't have to use the latter if you want all the glue:

```js
const win = new BrowserWindow({show: false})
const web = win.getContentView()
win.setContentView(new ImageView())
win.show()

win.webContents.once('did-finish-load', win.setContentView(web))
win.loadURL('https://github.com')
```

Adding a `WebContentsView` to `TopLevelWindow` won't automatically setup the glue code, because users would be able to add multiple child views in future so it is not certain which view would be the major client area.

And yeah it is very reasonable to document all the behaviors of `BrowserWindow`, and we should ensure users can replicate the behaviors using our public APIs. Currently the implementation of `BrowserWindow` has already been separated from `TopLevelWindow`, so you can read its source code to have an idea of what it is doing:
https://github.com/electron/electron/blob/master/lib/browser/api/browser-window.js
https://github.com/electron/electron/blob/master/shell/browser/api/electron_api_browser_window.cc

zcbenz:
@electron/wg-api I have updated the spec:

* Remove `webContents.create()`
* Make the constructor of `WebContentsView` take `webPreferences` instead of `WebContents`
* Add `TopLevelWindow.getContentView()` (Required for implementing a splash screen for `BrowserWindow`)
* Add `WebContentsView.webContents` property (Required for refactoring `WebContentsView` to take `webPreferences`)

Also on the behavior when a View with a parent is added to another parent, both Cocoa and `views` library would remove the View from the old parent and then add it to the new parent, I think we can just do the same for the `setContentView` API.

wg-api/spec-documents/views-api-1-content-view.md
```diff
@@ -0,0 +1,99 @@
...
+  const win = new TopLevelWindow({})
+  const imageView = new ImageView()
+  imageView.setImage(__dirname + '/splash.png')
+  win.setContentView(imageView)
+
+  const contents = webContents.create({})
+  const web = new WebContentsView(contents)
+  contents.once('did-finish-load', () => win.setContentView(web))
+  contents.loadURL('https://github.com/')
+})
```
jkleinsc:
```suggestion
  const web = new WebContentsView({})
  web.webContents.once('did-finish-load', () => win.setContentView(web))
  web.webContents.loadURL('https://github.com/')
})
```

itsananderson:
> * Remove `webContents.create()`

When you say "remove" do you mean remove from this spec, or remove from Electron entirely? `webContents.create()` [already exists](https://github.com/electron/electron/blob/07cd70a37e62a0203e5d2cc9b2e25f04363a94cc/shell/browser/api/electron_api_web_contents.cc#L2842), though I think it's not documented.

zcbenz:
> The only thing I couldn't figure out is why or where `TopLevelWindow.getContentView()` is needed?

It would be required if user wants to keep using `BrowserWindow` and temporarily swap its content view with an ImageView to implement splash screen. You can check the example in the comment above: https://github.com/electron/governance/pull/254#issuecomment-601474215.

It would also be required to write tests on how `setContentView` affects `BrowserWindow`.

> > * Remove `webContents.create()`
> 
> When you say "remove" do you mean remove from this spec, or remove from Electron entirely? `webContents.create()` [already exists](https://github.com/electron/electron/blob/07cd70a37e62a0203e5d2cc9b2e25f04363a94cc/shell/browser/api/electron_api_web_contents.cc#L2842), though I think it's not documented.

I mean we won't make the API public as part of this spec. This API is already being used heavily in our tests.

PR #282, opened 2020-05-08
chore: remove self from workgroup membership

ckerr:
Removing myself from the WG member list.

There's no underlying issue here; I just haven't been spending any time on Security WG topics for awhile now, so this change seems appropriate.

CC @electron/wg-security @MarshallOfSound @nornagon


nornagon:
You're welcome back any time!

PR #311, opened 2020-07-08
chore: QA WG OKR records

sofianguy:


sofianguy:
Closing due to sunsetting QA WG

PR #315, opened 2020-07-23
docs: release notes for July

codebytere:
Adds some missing release notes for July.


sofianguy:
Screen Shot 2020-07-23 at 1 07 57 PM

thanks @codebytere. are we missing notes from the first two weeks of July?

codebytere:
One week we did not have a meeting (cancelled on arrival) and the other week I fear may have accidentally been deleted :(

PR #328, opened 2020-08-24
docs: amend the Observer role to be more generalized

erickzhao:
Very open to changing the terms I used to describe the two parties involved.


policy/observers.md
```diff
@@ -0,0 +1,22 @@
+# Observers
+
+**Administrator:** A Working Group in Electron Governance or a company represented in a Working Group wishing to introduce an Observer into their Working Group.
+
+**Temporary worker:** A person employed or otherwise contracted by an Administrator for a temporary and fixed duration. Their role must entail that they primarily work on the Electron framework and/or other responsibilities of the Working Group they wish to join.
+
+
+## Working Group Membership
```
sofianguy, line 8: I think we should specify where the contractual date can be located. @codebytere had an example: "this role must have a predetermined start and end date, noted in their section of the WG membership list"

policy/observers.md
```diff
@@ -0,0 +1,22 @@
+# Observers
+
+**Administrator:** A Working Group in Electron Governance or a company represented in a Working Group wishing to introduce an Observer into their Working Group.
```
sofianguy, line 3: I think instead of `Administrator`, we can use either `maintainer` or `working group` as defined in https://github.com/electron/governance#definitions

policy/observers.md
```diff
@@ -0,0 +1,22 @@
...
+
+**Temporary worker:** A person employed or otherwise contracted by an Administrator for a temporary and fixed duration. Their role must entail that they primarily work on the Electron framework and/or other responsibilities of the Working Group they wish to join.
```
sofianguy, line 5: Suggestions: 
* Observer (what the interns used last year) 
Screen Shot 2020-08-24 at 3 53 23 PM

* Collaborator (see definition https://github.com/electron/governance#definitions)

erickzhao: What about "Temporary collaborator"? I'm hesitant to use either of those directly because there should be a distinction between the role they need to qualify for Observership and the Observer role itself. Likewise, the existing collaborator definition doesn't really encompass what we need here.

policy/observers.md
erickzhao: Any suggestion on how to reword this with the existing definitions to capture the same stipulations?

policy/observers.md
sofianguy: Great point. I'd be good with "Temporary Collaborator"

policy/observers.md
sofianguy: I take it back. I think it's fine as Administrator after rereading the whole doc it makes sense.

erickzhao:
@groundwater 

You're right about the potential "Administrator" mixup. I've changed all references to "Sponsor" instead.

> I think voting is not a chartered concept, but something defined by the WGs themselves. We may want to be more general, such as observers may not have complete rights and permissions of full WG members.

Good point. I don't think I've ever voted in Outreach/Ecosystem beyond a very informal temperature check. I'll amend this point and tag you to re-review.

jkleinsc (approved):

groundwater (approved):

PR #357, opened 2020-10-19
docs: tweak text of issue triage playbook responses

ckerr:
I'm on triage duty this week and some of these responses seemed like they needed a refresh -- e.g. improving phrasing, removing commasplices, or pointing users in the right direction. The only unifying theme across the changes is that they were needed by some of the issues that I triaged :smile_cat: and PRing them separately seemed like overkill.

CC @electron/wg-releases

### DIFF
--- a/playbooks/responses/blocked-needs-repro.md
+++ b/playbooks/responses/blocked-needs-repro.md
@@ -1,26 +1,28 @@
 ## If No Repro Included
 
-Thank you for taking the time to report this issue and helping to make Electron better.
+Thanks for reporting this and helping to make Electron better!
 
-Would it be possible for you to fork [electron-quick-start](https://github.com/electron/electron-quick-start) to make a stand-alone app that reproduces this issue?
+Would it be possible for you to make a standalone testcase with only the code necessary to reproduce the issue? For example, [Electron Fiddle](https://www.electronjs.org/fiddle) is a great tool for making small test cases and makes it easy to publish your test case to a [gist](https://gist.github.com) that Electron maintainers can use.
 
 Stand-alone test cases make fixing issues go more smoothly: it ensure everyone's looking at the same issue, it removes all unnecessary variables from the equation, and it can also provide the basis for automated regression tests.
 
-I'm adding the `blocked/needs-repro` label for this reason. After you make a test case (a [gist](https://gist.github.com) is preferred since gists can be loaded directly into [Electron Fiddle](https://www.electronjs.org/fiddle) for testing), please link to it in a followup comment.
+I'm adding the `blocked/needs-repro` label for this reason. After you make a test case, please link to it in a followup comment.
 
 Thanks in advance! Your help is appreciated.
 
+
+
 ## If Repro Requires Third-Party Tooling
 
-Thank you for taking the time to report this issue and helping to make Electron better.
+Thanks for reporting this and helping to make Electron better!
 
-Researching issues that depend on third-party code is usually not feasible for the Electron team due to the time that would be required to research reported issues with open-ended dependencies.
+Because of time constraints, triaging code with third-party dependencies is usually not feasible for a small team like Electron's.
 
-Would it be possible for you to fork [electron-quick-start](https://github.com/electron/electron-quick-start) for a small app that reproduces the issue by itself without other code?
+Would it be possible for you to make a standalone testcase with only the code necessary to reproduce the issue? For example, [Electron Fiddle](https://www.electronjs.org/fiddle) is a great tool for making small test cases and makes it easy to publish your test case to a [gist](https://gist.github.com) that Electron maintainers can use.
 
 Stand-alone test cases make fixing issues go more smoothly: it ensure everyone's looking at the same issue, it removes all unnecessary variables from the equation, and it can also provide the basis for automated regression tests.
 
-I'm adding the `blocked/needs-repro` label for this reason. After you make a test case (a [gist](https://gist.github.com) is preferred since gists can be loaded directly into [Electron Fiddle](https://www.electronjs.org/fiddle) for testing), please link to it in a followup comment.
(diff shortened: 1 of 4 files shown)


playbooks/responses/blocked-needs-repro.md
```diff
@@ -2,24 +2,25 @@
...
+Researching issues that depend on third-party code is usually not feasible for the Electron team due to the time that would be required to research reported issues with open-ended dependencies.
+
+Would it be possible for you to fork [electron-quick-start](https://github.com/electron/electron-quick-start) for a small app that reproduces the issue by itself without other code?
 
 Stand-alone test cases make fixing issues go more smoothly: it ensure everyone's looking at the same issue, it removes all unnecessary variables from the equation, and it can also provide the basis for automated regression tests.
 
-I'm adding the `blocked/needs-repro` label for this reason. After you've responded, please link to the test case and @ me in a followup comment.
+I'm adding the `blocked/needs-repro` label for this reason. After you make a test case (a [gist](https://gist.github.com) is preferred since gists can be loaded directly into [Electron Fiddle](https://www.electronjs.org/fiddle) for testing), please link to it in a followup comment.
```
codebytere, line 23: if we're asking users to fork quick-start, it's somewhat confusing to ask them to make a gist in the same response - we should probably stick to just one

playbooks/responses/end-of-life.md
```diff
@@ -1,9 +1,7 @@
-Thank you for taking the time to report this issue and helping to make Electron better.
+Thank you for reaching out!
 
-This issue has been marked for closing because this series of Electron has reached end-of-life and is no longer supported.
+The version of Electron reported in this issue has reached end-of-life and is no longer supported, so this issue will be closed.
 
-If you're still experiencing this issue in the latest version of Electron and would like this issue to remain open, please add a comment specifying the version you're testing with and any other information that a maintainer needs to know to reproduce the issue.
-
-I'm setting the `blocked/need-info` label for the above reasons.
+If you're still experiencing this issue on a supported version of Electron and would like this issue to remain open, please file a new issue with that version of Electron.
```
codebytere, line 5: +1, i think this shifts the onus to them in a good way.

playbooks/responses/blocked-needs-repro.md
```diff
@@ -2,24 +2,25 @@
 
 Thank you for taking the time to report this issue and helping to make Electron better.
 
-Would it be possible for you to fork [electron-quick-start](https://github.com/electron/electron-quick-start) for a small app that reproduces the issue by itself?
+Would it be possible for you to fork [electron-quick-start](https://github.com/electron/electron-quick-start) to make a stand-alone app that reproduces this issue?
 
 Stand-alone test cases make fixing issues go more smoothly: it ensure everyone's looking at the same issue, it removes all unnecessary variables from the equation, and it can also provide the basis for automated regression tests.
 
-I'm adding the `blocked/needs-repro` label for this reason. After you've responded, please link to the test case and @ me in a followup comment.
+I'm adding the `blocked/needs-repro` label for this reason. After you make a test case (a [gist](https://gist.github.com) is preferred since gists can be loaded directly into [Electron Fiddle](https://www.electronjs.org/fiddle) for testing), please link to it in a followup comment.
```
malept, line 9: I'm not sure if linking to GitHub Gists is useful here - we probably want to emphasize the use of Fiddle to save the Gist, as opposed to them trying to do it by hand.

playbooks/responses/not-an-issue.md
```diff
@@ -1 +1,3 @@
-GitHub issues are for feature requests and bug reports, questions about using Electron or code assistance requests should be directed to the [community](https://github.com/electron/electron#community).
+The Electron team uses its issues for tracking feature requests and bug reports.
+
+Please visit https://github.com/electron/electron#community for a list of resources that are better oriented for questions and help. Thank you!
```
malept, line 3: :grimacing: the [transitive link](https://github.com/electron/electron/blob/v12.0.0-nightly.20201021/docs/tutorial/support.md) here lists the Atom Slack, not the Discord server...

We should fix that, but FWIW I've been using this link instead:
```suggestion
Please visit https://www.electronjs.org/community for a list of resources that are better oriented for questions and help. Thank you!
```

playbooks/responses/blocked-needs-repro.md
ckerr: Hmm, I agree we don't want to make it confusing.

I _really_ like the idea of pushing users towards gists for their repro code since it is just so _efficient_ for testing... and I also like the idea of providing a good starting point a la quick-start to make the repro cases more uniform.

I'm wondering if there is some easy-to-understand way to do both of these things?

playbooks/responses/blocked-needs-repro.md
ckerr: Hmm you may be right. I intentionally avoided that step because installing Fiddle could be seen as burdensome to a bug reporter; but then again, you could make the same argument against gists. And if we recommend Fiddle we'll get gist support built-in, while the reverse is not the case.

@codebytere commented above that endorsing electron-quick-start _and_ gists might be too confusing as well. Maybe Fiddle is the answer to all of this, since we would just recommend fiddle and that tool automatically points people in the right direction.

playbooks/responses/not-an-issue.md
ckerr: Link fixed in cdbc8549daa5107d9d18f256bfb5e53cce65eef9.

playbooks/responses/blocked-needs-repro.md
ckerr: Update: I've revised this again in a26ffd95ff68d1725a6655b2fad81fe3189c0979 based on feedback from @codebyte and @malept

codebytere (approved):

malept (approved):

PR #359, opened 2020-10-20
chore: update Outreach WG OKRs for 2020

erickzhao:
Back in August, we had a meeting to update our OKRs since we didn't finalize them during the Electron Maintainer Summit. We generated an Objective and a few Initiatives, but never updated our working group's public README.

Here we go, 2 months later!

Miro board, for reference: https://miro.com/app/board/o9J_knuYqMI=/

To review: we also had "Measures" written down on the board. Do we want these to be on the README as well? cc @groundwater @molant 

As a note, we also had an action item to write a blog post about the output from the Summit, but I think that ship has sailed at this point. Maybe next time?

### DIFF
--- a/wg-outreach/README.md
+++ b/wg-outreach/README.md
@@ -17,12 +17,12 @@ Grows the Electron community
 ## Current Objective and Key Results
 **Objective:**
 
-Electron is easy for new people to approach and make meaningful contributions.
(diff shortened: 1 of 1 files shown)


sofianguy (approved):
Thanks Erick!

codebytere (approved):

PR #360, opened 2020-10-22
chore: Join admin, leave outreach & ecosystem

felixrieseberg:
With my new role at Slack comes a move at Electron, too -- and I'm joining `wg-admin` to take over the baton from Slack's previous members in that working group.

We've always said that members of `wg-admin` shouldn't have a vote in any of the operative working groups, which I believe is the correct principle to follow. The way I see it, I can still work on `fiddle` and contribute to various outreach projects, I'll just need to follow what those two working groups think. With that in mind, I'm not going anywhere - I'm just swapping out my two voting positions for another one.

(Also, we've discussed that @molant is now the chair of `wg-admin`)


erickzhao (approved):
😭 🥳

vhashimotoo (approved):

ckerr (approved):
Approved and I see GitHub thinks this is ready to merge --

to be pedantic though, we really need signoff from admin for the `wg-administrative/*` changes. Pinging @groundwater since he's the only person in both the before & after :slightly_smiling_face:

groundwater (approved):

PR #368, opened 2020-11-02
docs: protocol api: simplification

dsanders11:
Simplification of the `protocol` module APIs.

>Expose `protocol.registerProtocol` and `protocol.interceptProtocol` to allow any type of protocol response, simplifying the API surface of the `protocol` module. Add several classes to replace the overloaded `ProtocolResponse` object. Deprecate the current `protocol.{intercept|register}{Buffer|File|Http|Stream|String}Protocol` methods.

[Rendered markdown](https://github.com/dsanders11/governance/blob/protocol-api-simplification/wg-api/spec-documents/protocol-simplification.md).

### DIFF
--- a/wg-api/spec-documents/protocol-simplification.md
+++ b/wg-api/spec-documents/protocol-simplification.md
@@ -1,0 +1,368 @@
+# Protocol Module Simplification
+
+## Summary
+
+Expose `protocol.registerProtocol` and `protocol.interceptProtocol` to allow
+any type of protocol response, simplifying the API surface of the `protocol`
+module. Add several classes to replace the overloaded
+`ProtocolResponse` object. Deprecate the current
+`protocol.{intercept|register}{Buffer|File|Http|Stream|String}Protocol`
+methods.
+
+## Platforms
+
+All
+
+## Impetus
+
+The current `protocol` module API is a bit unwieldy with 10 typed
+methods. They all use an overloaded [`ProtocolResponse`] object and the
+existing implementation silently ignores keys not used by any given method.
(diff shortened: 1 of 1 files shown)


nornagon:
I like this general proposal but I think we can go further! `intercept` and `register` don't need to be two different things; as the only difference between them is that `register` is for protocols that don't exist yet and `intercept` is for ones which do, which is something we can determine automatically.

I think we should just have one method, `protocol.handle` (thanks @dsanders11 for the name suggestion), that can either intercept an existing protocol or register a new one. Multiple handlers registered for the same protocol can chain. For example,

```js
protocol.handle("app", async (request, next) => {
  const response = await getResponseFor(request)
  return response ?? next()
})
```

Here `next` would delegate to whatever existing handler there was, or else to some default failure behavior if there was no existing handler.

```js
protocol.handle("http", async (request, next) => {
  if (request.path === '/')
    // I'm unsure if this should be something like "new StringResponse", or whether
    // a plain object is better
    return {content: '<body>hello world', mime: 'text/html'}
  else if (request.path === '/some/image.png')
    return {file: '/path/to/my/image.png'}
  else if (request.path === '/some/video.mp4')
    return {content: generateVideoStream(), mime: 'video/mp4'}
  else if (request.path === '/sneakily-cross-origin')
    return {url: 'https://foo.bar/baz'}
  else return next()
})
```

samuelmaddock:
@nornagon has this design landed?

dsanders11:
@samuelmaddock, `protocol.handle` landed which addresses most (all?) of the use cases from this PR, so I'll close this.

PR #404, opened 2021-02-22
chore: de-commission and Archive QA Working Group

groundwater:
The QA Working Group has requested to disband due to a recent lack of sufficient participation.

If interest were to change in the future, the group could be re-chartered. It is better in the mean time to close the group if it's not going to perform a regular function.

QA working group members who are not part of another working group should be converted to non-governance memberships. This would include Slack and GitHub permissions. I am in favor of giving folks a 2 week grace period for them to join another working group to maintain their governance membership.

I chose to explicitly place the working group in an `archive` directory versus deleting. I realize git history makes everything searchable, but an explicit archive directory is a little more user-friendly IMO.


felixrieseberg (approved):
Thank you for taking care of this :bow:

Madsadglad (approved):

Madsadglad (approved):
😖😭😃

jkleinsc (approved):

codebytere (approved):

codebytere:
Just needs semantic commit and then i think we can merge :)

felixrieseberg:
Looks good: One request: Let's keep the channel alive until @MariaBeliaeva had a chance to reach out to the former WG members so that they are informed and may join another working group.

SlackedMarcel:
The community (wg-qa in slack) has been notified of the change, and instructions to join another group was shared. We can move forward with this PR.

ssabawissabawi (approved):

PR #447, opened 2021-09-09
chore: update Releases chair to @codebytere

codebytere:
As in title


clavin (approved):
super micro nit: the space before `**Chair**` got lost, but it doesn't really change anything & can be fixed next time chair rotates

VerteDinde (approved):

PR #454, opened 2021-11-03
docs: add spec for clipboard rearchitecture

codebytere:
This document outlines a proposal for the future of the `clipboard` module provided by Electron.

Over the course of the last few months, Chromium has been making significant changes to their Clipboard API, including the ability to handle [Custom Formats](https://bugs.chromium.org/p/chromium/issues/detail?id=106449). Electron has had to break small aspects of its clipboard API to account for this, and it's unclear what other changes we may be forced to adapt to as they continue this work. Beyond this, users have been asking for an improved Clipboard API as well as surfaced a number of bugs with its current implementation - the most detailed of which is [here](https://github.com/electron/electron/issues/23156). Given this, @jkleinsc and i have put together a more streamlined proposal for its future.

### DIFF
--- a/wg-api/spec-documents/clipboard-rearchitecture.md
+++ b/wg-api/spec-documents/clipboard-rearchitecture.md
@@ -1,0 +1,105 @@
+# Clipboard Module Rearchitecture
+
+> Shelley Vohr and John Kleinschmidt
+
+## Summary
+
+This RFC outlines a rearchitecturing of the `clipboard` module in Electron to improve alignment with the Clipboard API as specified by the [W3C](https://w3c.github.io/clipboard-apis/#clipboard-interface).
+
+## Platforms
+
+All - macOS, Windows, Linux.
+
+## Impetus
+
+Over the course of the last few months, Chromium has been making significant changes to their Clipboard API, including the ability to handle [Custom Formats](https://bugs.chromium.org/p/chromium/issues/detail?id=106449). Electron has had to break small aspects of its clipboard API to account for this, and it's unclear what other changes we may be forced to adapt to as they continue this work. Beyond this, users have been asking for an improved Clipboard API as well as surfaced a number of bugs with its current implementation - the most detailed of which is [here](https://github.com/electron/electron/issues/23156).
+
+Some other related issues include:
+
+- [#31130](https://github.com/electron/electron/issues/31130)
+- [#31115](https://github.com/electron/electron/issues/31115)
+- [#30699](https://github.com/electron/electron/issues/30699)
+- [#26377](https://github.com/electron/electron/issues/26377)
+- [#11838](https://github.com/electron/electron/issues/11838)
+- [#9035](https://github.com/electron/electron/issues/9035)
+- [#5078](https://github.com/electron/electron/issues/5078)
+- [electron/governance#229](https://github.com/electron/governance/pull/229)
+
+Electron provided a `clipboard` API to its users preceding the specification of the [Clipboard API by the W3C](https://w3c.github.io/clipboard-apis/#clipboard-interface). As a result, we locked into an API surface and created an API contract with our users that became progressively harder to alter without breaking said contract while the web drove a more standardized approach. On top of this confusion, the `clipboard` module provided by Electron is a dual-process module, meaning that it exists in parallel to the more robust API provided by Chromium in the renderer process via `navigator.clipboard` global. 
+
+At this point, it becomes clear that it is in users' and maintainers' best interest to reconsider how we approach the `clipboard` module - breaking changes might be difficult in the short term but provide more consistency and functionality as we look to the future of the module.
+
+## API Design
+
+The largest consideration for changes to the `clipboard` module is to what degree we should choose to align with the W3C specification. As we consider this, we should consider what needs or use cases might exist for Electron uses which necessarily wouldn't exist for Chrome users, and how we would plan to account for that in our new design.
+
+### Potential Options
+
+1. Strict alignment with W3C specification - no additional APIs exposed.
+2. Alignment with W3C where possible, with some additional APIs exposed to handle desktop use cases which aren’t addressed by the specification.
+3. Leave API as-is and continue to adapt it as required by upstream changes by Chromium development.
+ 
+### Preferred Approach/Solution
+
+Alignment with W3C where possible, with some additional APIs exposed to handle desktop use cases which aren’t considered by the specification.
+
+**APIs to Preserve**
+* `clipboard.readBookmark()`
+  * Not supported through existing Web APIs or in the specification.
+* `clipboard.writeBookmark(title, url[, type])`
+  * Not supported through existing Web APIs or in the specification.
+* `clipboard.readFindText()` _macOS_
+  * Not supported through existing Web APIs or in the specification.
+* `clipboard.writeFindText(text)`
+  * Not supported through existing Web APIs or in the specification.
+* `clipboard.clear([type])`
+  * Not supported through existing Web APIs or in the specification.
+* `clipboard.readText([type])`
+  * Already implemented per specification.
+* `clipboard.writeText(text[, type])`
+  * Already implemented per specification.
+
+**APIs to Remove**
+* `clipboard.availableFormats([type])`
+  * Superseded by `clipboard.read` [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read)
+* `clipboard.has(format[, type])` 
+  * Superseded by `clipboard.read` [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read)
+* `clipboard.readBuffer(format)`
+  * Superseded by `clipboard.read`[Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read)
+* `clipboard.writeBuffer(format, buffer[, type])`
+  * Superseded by `clipboard.write` [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write)
+* `clipboard.readHTML()`
+  * Superseded by `clipboard.read` with `text/html` MIME type
+* `clipboard.writeHTML(markup[, type])`
+  * Superseded by `clipboard.write` with `text/html` MIME type
+* `clipboard.readImage([type])`
+  * Superseded by `clipboard.read()` with `image/*` (where * is bmp, gif, jpeg, png, svg+xml, etc.)
+* `clipboard.writeImage(image[, type])`
+  * Superseded by `clipboard.write`  with `image/*` (where * is bmp, gif, jpeg, png, svg+xml, etc.)
+* `clipboard.readRTF([type])`
+  * Superseded by `clipboard.read` with `application/rtf` MIME type
+* `clipboard.writeRTF(text[, type])`
+  * Superseded by `clipboard.write` with `application/rtf` MIME type
+
+**APIs to Modify**
+* `clipboard.read(format)`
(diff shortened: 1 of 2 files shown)


wg-api/spec-documents/clipboard-rearchitecture.md
```diff
@@ -0,0 +1,105 @@
...
+* `clipboard.has(format[, type])` 
+  * Superseded by `clipboard.read` [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read)
+* `clipboard.readBuffer(format)`
+  * Superseded by `clipboard.read`[Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read)
+* `clipboard.writeBuffer(format, buffer[, type])`
+  * Superseded by `clipboard.write` [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write)
+* `clipboard.readHTML()`
+  * Superseded by `clipboard.read` with `text/html` MIME type
+* `clipboard.writeHTML(markup[, type])`
+  * Superseded by `clipboard.write` with `text/html` MIME type
```
ckerr, line 74: Does this have the same semantics as clipboard.write with 'text/html'? The docs for these two functions imply there's an internal transform, e.g. the addition of a `<meta charset='utf-8'>`

This question might not matter since this proposal is a breaking change anyway, but I'm curious so I'll ask anyway :smile:

wg-api/spec-documents/clipboard-rearchitecture.md
```diff
@@ -0,0 +1,105 @@
...
+* `clipboard.readImage([type])`
+  * Superseded by `clipboard.read()` with `image/*` (where * is bmp, gif, jpeg, png, svg+xml, etc.)
+* `clipboard.writeImage(image[, type])`
+  * Superseded by `clipboard.write`  with `image/*` (where * is bmp, gif, jpeg, png, svg+xml, etc.)
```
ckerr, line 78: We would still be able to pass native images in through clipboard.read() & .write(), correct?

wg-api/spec-documents/clipboard-rearchitecture.md
```diff
@@ -0,0 +1,105 @@
...
+
+**APIs to Modify**
+* `clipboard.read(format)`
+  * Modify to bring into spec compliance with [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read) `clipboard.read`
+  * Ensure that raw formats are preserved
+    * Custom MIME types e.g. `electron/filepath`
+* `clipboard.write(data[, type])​`
+  * Modify to bring into spec compliance with [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read) `clipboard.read`
+  * Ensure that raw formats are preserved
+    * Custom MIME types e.g. `electron/filepath`
```
ckerr, line 92: It looks like there's a muddling of the meaning of 'type' here -- in `main` the type is clipboard | selection; here, it's mime-type.

If this draft was implemented, would Electron lose the ability to read/write from the Linux selection buffer?

wg-api/spec-documents/clipboard-rearchitecture.md
```diff
@@ -0,0 +1,105 @@
...
+  
+**APIs to Add**
+* `onclipboardchange` event
```
ckerr, line 95:
```suggestion
* `clipboardchange` event
```

Fixing the name to [match the spec](https://w3c.github.io/clipboard-apis/#clipboard-event-clipboardchange)

wg-api/spec-documents/clipboard-rearchitecture.md
```diff
@@ -0,0 +1,105 @@
...
+
+**APIs to Preserve**
+* `clipboard.readBookmark()`
+  * Not supported through existing Web APIs or in the specification.
+* `clipboard.writeBookmark(title, url[, type])`
+  * Not supported through existing Web APIs or in the specification.
+* `clipboard.readFindText()` _macOS_
+  * Not supported through existing Web APIs or in the specification.
+* `clipboard.writeFindText(text)`
+  * Not supported through existing Web APIs or in the specification.
```
ckerr, line 54: Could these be absorbed into read() / write() via the custom `electron/` mime types that are mentioned below in the 'APIs to Modify' section?

find text probably could, as it's a string. I'm less sure about bookmarks

ckerr (commented):
Overall

wg-api/spec-documents/clipboard-rearchitecture.md
zcbenz: On macOS there is no event-based API to monitor clipboard change, the only way to implement this event is to check clipboard with an interval. We might need another API to control the interval.

wg-api/spec-documents/clipboard-rearchitecture.md
```diff
@@ -0,0 +1,105 @@
...
+  * Superseded by `clipboard.read()` with `image/*` (where * is bmp, gif, jpeg, png, svg+xml, etc.)
+* `clipboard.writeImage(image[, type])`
+  * Superseded by `clipboard.write`  with `image/*` (where * is bmp, gif, jpeg, png, svg+xml, etc.)
+* `clipboard.readRTF([type])`
+  * Superseded by `clipboard.read` with `application/rtf` MIME type
+* `clipboard.writeRTF(text[, type])`
+  * Superseded by `clipboard.write` with `application/rtf` MIME type
+
+**APIs to Modify**
+* `clipboard.read(format)`
```
zcbenz, line 85: The Web API's `Clipboard.read` method returns an array of `ClipboardItem`, which includes the type information of the clipboard data, so the `format` parameter is not needed.

wg-api/spec-documents/clipboard-rearchitecture.md
```diff
@@ -0,0 +1,105 @@
...
+  * Not supported through existing Web APIs or in the specification.
+* `clipboard.readText([type])`
+  * Already implemented per specification.
+* `clipboard.writeText(text[, type])`
+  * Already implemented per specification.
+
+**APIs to Remove**
+* `clipboard.availableFormats([type])`
+  * Superseded by `clipboard.read` [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read)
+* `clipboard.has(format[, type])`
```
zcbenz, line 65: Having a `has` API is probably still needed, otherwise one has to use `read` to get all types of data at once, and the literate them to find the needed one.

wg-api/spec-documents/clipboard-rearchitecture.md
```diff
@@ -0,0 +1,105 @@
...
+1. Strict alignment with W3C specification - no additional APIs exposed.
+2. Alignment with W3C where possible, with some additional APIs exposed to handle desktop use cases which aren’t addressed by the specification.
+3. Leave API as-is and continue to adapt it as required by upstream changes by Chromium development.
+ 
+### Preferred Approach/Solution
+
+Alignment with W3C where possible, with some additional APIs exposed to handle desktop use cases which aren’t considered by the specification.
+
+**APIs to Preserve**
+* `clipboard.readBookmark()`
```
zcbenz, line 47: I think this API can also be replaced by `read` with custom MIME type?

zcbenz (commented):
The spec should also document the type (i.e. `ClipboardItem`) being passed between `read` and `write` APIs.

wg-api/spec-documents/clipboard-rearchitecture.md
jkleinsc: yes

wg-api/spec-documents/clipboard-rearchitecture.md
jkleinsc: @ckerr the `type` here is the current argument we have to use the Linux selection buffer.

wg-api/spec-documents/clipboard-rearchitecture.md
jkleinsc: Bookmark is a little weird because it is two items - a title and a url that would need to be consolidated into one item, but we could accommodate via JSON encoding.

wg-api/spec-documents/clipboard-rearchitecture.md
ckerr: Got it. I was misunderstanding how the W3C spec denoted mime-types.

wg-api/spec-documents/clipboard-rearchitecture.md
```diff
@@ -0,0 +1,285 @@
...
+
+**APIs to Modify**
+* `clipboard.read([clipboardType])`
+  * Modify to bring into spec compliance with [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read) `clipboard.read`
+  * Ensure that raw formats are preserved
+    * Custom MIME types e.g. `electron/filepath`
+  * clipboardType is only used for Linux to specify if clipboard is regular clipboard or `selection` clipboard.
+* `clipboard.write(data[, clipboardType]])​`
+  * Modify to bring into spec compliance with [Web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read) `clipboard.read`
+  * Ensure that raw formats are preserved
```
codebytere, line 92: Burndown output - we shouldn't use mystery mime types for types which represent location as opposed to content type. We should instead add Electron-specific clipboard-ish functionality as discrete methods alongside coalescing existing MIME types into methods aligned with the Web APIs.

jkleinsc: Using custom mime types corresponds to what the web is doing via custom clipboard 
 types: https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md

wg-api/spec-documents/clipboard-rearchitecture.md
jkleinsc: `has` can be replicated with a fairly short amount of JS:
```js
async function has(format, clipboardType) {
  const clipboardItems = await clipboard.read(clipboardType);
  const matchingClipboardItem = clipboardItems.find(clipboardItem => {
    return clipboardItem.types.includes(format);
  });
  if (matchingClipboardItem) {
    return true;
  }
}
```

codebytere: I'd be in favor of keeping `has` as well - I'd probably also be in favor of leaving it in cpp given the vast bulk of clipboard will be implemented at that layer and it's likely marginally faster.

wg-api/spec-documents/clipboard-rearchitecture.md
codebytere: cc @nornagon iirc you strongly preferred avoiding custom mime types - what are your thoughts given the above? I'd also like to choose vaguely the same approach as upstream, at minimum for future-proofing purposes

nornagon: Honestly I don't remember what my objection was; following the Web standard seems like a reasonable thing to do. Notably, the Web standard looks like it's only allowing _very circumscribed_ custom formats, in the pattern `web foo/bar`. Which I don't think was what this proposal was outlining.

wg-api/spec-documents/clipboard-rearchitecture.md
jkleinsc: @ckerr I just checked.  Electron calls the underlying Chromium APIs, so that `<meta charset='utf-8'>` gets added by chromium and will continue to do so under the new implementation.  You can test it yourself at https://fuzzy-wobbly-splash.glitch.me/

jkleinsc:
Closing in favor of https://github.com/electron/rfcs/pull/19

PR #478, opened 2022-03-22
docs: scroll-view proposal

ddramone:
Proposing API changes required for adding a native ScrollView.

Link to the initial feature request: https://github.com/electron/electron/issues/32751

### DIFF
--- a/wg-api/spec-documents/scroll-view.md
+++ b/wg-api/spec-documents/scroll-view.md
@@ -48,7 +48,10 @@ win.addChildView(scroll)
 
 ## API Design
 
-In order to achieve the example above, we've made following changes to the API
+- `View` - The base class for native UI elements.
+- `ImageView` - The native UI element that displays image.
+- `WebContentsView` - The native UI element that displays the content of `WebContents`.
+- `ScrollView` - The native UI element that make any View scrollable.
 
 ## View
 
--- a/wg-api/spec-documents/scroll-view.md
+++ b/wg-api/spec-documents/scroll-view.md
@@ -247,7 +247,7 @@ Returns:
 
 * `event` Event
 
-Emitted at the beginning of user-initiated scroll tracking.
+Emitted at the beginning of user-initiated scrolling.
 
 #### Event: 'end-scroll' _Experimental_
 
@@ -255,11 +255,11 @@ Returns:
 
 * `event` Event
 
-Emitted at the end of scroll tracking.
+Emitted at the end of user-initiated scrolling.
 
 ## Rollout Plan
 
 - add the documentation to [Views API (Part 1)](views-api-1-content-view.md)
-- Move the implementation from `BrowserView` to `InspectableWebContentsView` (in particular the implementation of draggable regions)
+- Move the implementation of `BrowserView`'s features (in particular the draggable regions) - to get them working properly in `WebContentsView`
 - hide all features (added by this spec) behind a command line switch
 
--- a/wg-api/spec-documents/scroll-view.md
+++ b/wg-api/spec-documents/scroll-view.md
@@ -79,11 +79,11 @@ Returns the position and size of the view.
 
 Change the background color of the view.
 
-#### `view.getParentView()`
+#### `view.getParent()`
 
 Returns `BaseView || null` - The parent view, otherwise returns `null`.
 
-#### `view.getParentWindow()`
+#### `view.getWindow()`
 
 Returns `BrowserWindow || null` - The window that the view belongs to, otherwise returns `null`.
 
--- a/wg-api/spec-documents/scroll-view.md
+++ b/wg-api/spec-documents/scroll-view.md
@@ -221,16 +221,15 @@ Returns `string` - The scroll view’s vertical scrolling elasticity mode.
 
 * `point` - The point in the `contentView` to scroll to.
 
-Scrolls the view’s closest ancestor `clipView` object so a point in the
-view lies at the origin of the clip view's bounds rectangle.
+Scroll to the horizontal (`point.x`) and vertical (`point.y`) position.
 
 #### `view.getScrollPosition()` _Experimental_
 
-Returns `Point`
+Returns `Point` - The horizontal and vertical scroll position.
 
 #### `view.getMaximumScrollPosition()` _Experimental_
 
-Returns `Point`
+Returns `Point` - The maximum horizontal and vertical scroll position.
 
 #### Events
 
--- a/wg-api/spec-documents/scroll-view.md
+++ b/wg-api/spec-documents/scroll-view.md
@@ -23,26 +23,26 @@ More details of the use-case and challenges related to workarounds: [#32751](htt
 ## Usage Example
 
 ```js
-const { BrowserView, BrowserWindow, BaseView, ScrollView, WebBrowserView } = require("electron");
+const { BrowserWindow, View, ScrollView, WebContentsView } = require("electron");
 
 const window = new BrowserWindow({ width: 1000, height: 500 })
 
 const scroll = new ScrollView()
 scroll.setBounds({ width: 1000, height: 500 })
 
-const scrollContent = new BaseView()
-scrollContent.setBounds({ width: 2000, height: 500 })
+const scrollContent = new View()
 
-const webBrowserView1 = new WebBrowserView()
-webBrowserView1.setBounds({ width: 1000, height: 500 })
-scrollContent.addChild(webBrowserView1)
+const webContentsView1 = new WebContentsView({})
+webContentsView1.setBounds({ width: 1000, height: 500 })
+scrollContent.addChildView(webContentsView1)
 
-const webBrowserView2 = new WebBrowserView()
-webBrowserView2.setBounds({ width: 1000, height: 500, x: 1000 })
-scrollContent.addChild(webBrowserView2)
+const webContentsView2 = new WebContentsView({})
+webContentsView2.setBounds({ width: 1000, height: 500, x: 1000 })
+scrollContent.addChildView(webContentsView2)
 
 scroll.setContentView(scrollContent);
-win.addChild(scroll);
+scroll.setContentSize({ width: 2000, height: 500 })
+win.addChildView(scroll)
 ```
 
 
@@ -50,18 +50,18 @@ win.addChild(scroll);
 
 In order to achieve the example above, we've made following changes to the API
 
-## BaseView
+## View
 
-A `BaseView` is a rectangle within the views View hierarchy. It is the base
-class for all all different views. 
+The base class for native UI elements. The current implementation has the
+ability to add child views. In this spec more abilities  are added.
 
 Process: **Main**
 
-`BaseView` is an EventEmitter
+`View` is an EventEmitter
 
-### `new BaseView()` _Experimental_
+### `new View()` _Experimental_
 
-Creates the new base view.
+Creates the new view.
 
 ### Instance Methods
 
@@ -91,24 +91,20 @@ Change the background color of the view.
 
 #### `view.getParent()` _Experimental_
 
-Returns `BaseView || null` - The parent view, otherwise returns `null`.
+Returns `View || null` - The parent view, otherwise returns `null`.
 
 #### `view.getWindow()` _Experimental_
 
-Returns `BrowserWindow || null` - The window that the view belongs to, otherwise returns `null`.
+Returns `BaseWindow || null` - The window that the view belongs to, otherwise returns `null`.
 
-#### `view.addChild(view)` _Experimental_
+#### `view.removeChildView(view)` _Experimental_
 
-* `view` BaseView
-
-#### `view.removeChild(view)` _Experimental_
-
-* `view` BaseView
+* `view` View
 
 #### `view.getChildren()` _Experimental_
 
-Returns `BaseView[]` - an array of all BaseViews that have been attached
-with `addChild`.
+Returns `View[]` - an array of all Views that have been attached
+with `addChildView`.
 
 ### Events
 
@@ -117,22 +113,22 @@ with `addChild`.
 Emitted when the view's size has been changed.
 
  
-## BrowserWindow 
+## BaseWindow 
 
 ### New instance methods
 
-#### `win.addChild(view)` _Experimental_
+#### `win.addChildView(view)` _Experimental_
 
-* `view` BaseView
+* `view` View
 
-#### `win.removeChild(view)` _Experimental_
+#### `win.removeChildView(view)` _Experimental_
 
-* `view` BaseView
+* `view` View
 
 #### `win.getChildren()` _Experimental_
 
-Returns `BaseView[]` - an array of all BaseViews that have been attached
-with `addChild`.
+Returns `View[]` - an array of all Views that have been attached
+with `addChildView`.
 
 
 ## ScrollView
@@ -142,7 +138,7 @@ any View scrollable. When the content is larger than the `ScrollView`,
 scrollbars will be optionally showed. When the content view is smaller
 then the `ScrollView`, the content view will be resized to the size of the
 `ScrollView`.
-It extends `BaseView`.
+It extends `View`.
 
 Process: **Main**
 
@@ -158,7 +154,7 @@ Set the contents. The contents is the view that needs to scroll.
 
 #### `scrollView.getContentView()` _Experimental_
 
-Returns `BaseView` - The contents of the `scrollView`.
+Returns `View` - The contents of the `scrollView`.
 
 #### `scrollView.setContentSize(size)` _Experimental_
 
@@ -191,7 +187,7 @@ Returns `string` - horizontal scrollbar mode.
 
 Returns `string` - vertical scrollbar mode.
 
-#### `scrollView.setHorizontalScrollElasticity(elasticity)` _macOS_ _Experimental_
+#### `scrollView.setHorizontalScrollElasticity(elasticity)` _Experimental_
 
 * `elasticity` string - Can be `automatic`, `none`, `allowed`. Default is `automatic`.
 
@@ -204,17 +200,17 @@ or the vertical scroller is hidden and the horizontal scroller is visible.
 * `none` - Disallow scrolling beyond document bounds on this axis.
 * `allowed` - Allow content to be scrolled past its bounds on this axis in an elastic fashion.
 
-#### `scrollView.setVerticalScrollElasticity(elasticity)` _macOS_ _Experimental_
+#### `scrollView.setVerticalScrollElasticity(elasticity)` _Experimental_
 
 * `elasticity` string - Can be `automatic`, `none`, `allowed`. Default is `automatic`.
 
 The scroll view’s vertical scrolling elasticity mode.
 
-#### `scrollView.getHorizontalScrollElasticity()` _macOS_ _Experimental_
+#### `scrollView.getHorizontalScrollElasticity()` _Experimental_
 
 Returns `string` - The scroll view’s horizontal scrolling elasticity mode.
 
-#### `scrollView.getVerticalScrollElasticity()` _macOS_ _Experimental_
+#### `scrollView.getVerticalScrollElasticity()` _Experimental_
 
 Returns `string` - The scroll view’s vertical scrolling elasticity mode.
 
@@ -259,23 +255,9 @@ Returns:
 
 Emitted at the end of scroll tracking.
 
-## WebBrowserView
-
-A `WebBrowserView` is the alternative for `BrowserView`. It extends `BaseView`, so it can be used as part of view's hierarchy.
-
-Process: **Main**
-
-### `new WebBrowserView([options])` _Experimental_
-
-* `options` Object (optional)
-  * `webPreferences` (optional)
-
-### Instance Properties
-
-#### `view.webContents` _Experimental_
-
-A `WebContents` object owned by this view.
-
 ## Rollout Plan
 
-TBD
+- add the documentation to [Views API (Part 1)](views-api-1-content-view.md)
+- Move the implementation from `BrowserView` to `InspectableWebContentsView` (in particular the implementation of draggable regions)
+- hide all features (added by this spec) behind a command line switch
+
--- a/wg-api/spec-documents/scroll-view.md
+++ b/wg-api/spec-documents/scroll-view.md
@@ -259,7 +259,8 @@ Emitted at the end of user-initiated scrolling.
 
 ## Rollout Plan
 
-- add the documentation to [Views API (Part 1)](views-api-1-content-view.md)
-- Move the implementation of `BrowserView`'s features (in particular the draggable regions) - to get them working properly in `WebContentsView`
-- hide all features (added by this spec) behind a command line switch
-
+- duplicate the BrowserView API (setAutoResize, setBounds etc.) into View API; targeting Electron v21
+- add the documentation to [Views API (Part 1)](views-api-1-content-view.md); targeting Electron v21
+- adjust BaseWindow API to support the hierarchy of views (WebContentsViews); targeting Electron v21
+- add the ScrollView API (as experimental); targeting Electron v22
+- optimization the ScrollView with WebContentsViews for performance; targeting Electron v22
--- a/wg-api/spec-documents/scroll-view.md
+++ b/wg-api/spec-documents/scroll-view.md
@@ -61,35 +61,35 @@ Process: **Main**
 
 `BaseView` is an EventEmitter
 
-### `new BaseView()`
+### `new BaseView()` _Experimental_
 
 Creates the new base view.
 
 ### Instance Methods
 
-#### `view.setBounds(bounds)`
+#### `view.setBounds(bounds)` _Experimental_
 
 Resizes and moves the view to the supplied bounds relative to its parent.
 
-#### `view.getBounds()`
+#### `view.getBounds()` _Experimental_
 
 Returns the position and size of the view.
 
-#### `view.setBackgroundColor(color)`
+#### `view.setBackgroundColor(color)` _Experimental_
 
 Change the background color of the view.
 
-#### `view.getParent()`
+#### `view.getParent()` _Experimental_
 
 Returns `BaseView || null` - The parent view, otherwise returns `null`.
 
-#### `view.getWindow()`
+#### `view.getWindow()` _Experimental_
 
 Returns `BrowserWindow || null` - The window that the view belongs to, otherwise returns `null`.
 
 ### Events
 
-#### Event: 'size-changed'
+#### Event: 'size-changed' _Experimental_
 
 Emitted when the view's size has been changed.
 
@@ -98,11 +98,15 @@ Emitted when the view's size has been changed.
 
 ### New instance methods
 
-#### `win.addChild(view)`
+#### `win.addChild(view)` _Experimental_
 
-#### `win.removeChild(view)`
+* `view` BaseView
 
-#### `win.getChildren()`
+#### `win.removeChild(view)` _Experimental_
+
+* `view` BaseView
+
+#### `win.getChildren()` _Experimental_
 
 Returns `BaseView[]` - an array of all BaseViews that have been attached
 with `addChild`.
@@ -114,17 +118,21 @@ A `ContainerView` can be used to embed additional views hierarchy into a `Browse
 
 Process: **Main**
 
-### `new ContainerView()`
+### `new ContainerView()` _Experimental_
 
 Creates the new container view.
 
 ### Instance Methods
 
-#### `containerView.addChild(view)`
+#### `containerView.addChild(view)` _Experimental_
 
-#### `containerView.removeChild(view)`
+* `view` BaseView
 
-#### `containerView.getChildren()`
+#### `containerView.removeChild(view)` _Experimental_
+
+* `view` BaseView
+
+#### `containerView.getChildren()` _Experimental_
 
 Returns `BaseView[]` - an array of all BaseViews that have been attached
 with `addChild`.
@@ -141,52 +149,54 @@ It extends `BaseView`.
 
 Process: **Main**
 
-### `new ScrollView()`
+### `new ScrollView()` _Experimental_
 
 Creates the new scroll view.
 
 ### Instance Methods
 
-#### `scrollView.setContentView(contents)`
+#### `scrollView.setContentView(contents)` _Experimental_
 
 Set the contents. The contents is the view that needs to scroll.
 
-#### `scrollView.getContentView()`
+#### `scrollView.getContentView()` _Experimental_
 
 Returns `BaseView` - The contents of the `scrollView`.
 
-#### `scrollView.setContentSize(size)`
+#### `scrollView.setContentSize(size)` _Experimental_
 
 Set the size of the contents.
 
-#### `scrollView.getContentSize()`
+#### `scrollView.getContentSize()` _Experimental_
 
 Returns the `size` of the contents.
 
-#### `scrollView.setHorizontalScrollBarMode(mode)`
+#### `scrollView.setHorizontalScrollBarMode(mode)` _Experimental_
 
-* `mode` string - Can be one of the following values: `disabled`, `hidden-but-enabled`, `enabled`. Default is `enabled`.
+* `mode` string - Can be `disabled`, `enabled-but-hidden`, `enabled`. Default is `enabled`.
 
 Controls how the horizontal scroll bar appears and functions.
-* `disable` - The scrollbar is hidden.
-* `hidden-but-enabled` - The scrollbar is hidden whether or not the contents are larger than the viewport, but the pane will respond to scroll events. This option is not available on MacOS.
+* `disabled` - The scrollbar is hidden.
+* `enabled-but-hidden` - The scrollbar is hidden whether or not the contents are larger than the viewport, but the pane will respond to scroll events. _Linux_  _Windows_
 *`enabled` - The scrollbar will be visible if the contents are larger than the viewport.
 
-#### `scrollView.setVerticalScrollBarMode(mode)`
+#### `scrollView.setVerticalScrollBarMode(mode)` _Experimental_
+
+* `mode` string - Can be `disabled`, `enabled-but-hidden`, `enabled`. Default is `enabled`.
 
 Controls how the vertical scroll bar appears and functions.
 
-#### `getHorizontalScrollBarMode()`
+#### `scrollView.getHorizontalScrollBarMode()` _Experimental_
 
 Returns `string` - horizontal scrollbar mode.
 
-#### `scrollView.getVerticalScrollBarMode()`
+#### `scrollView.getVerticalScrollBarMode()` _Experimental_
 
 Returns `string` - vertical scrollbar mode.
 
-#### `scrollView.setHorizontalScrollElasticity(elasticity)` _macOS_
+#### `scrollView.setHorizontalScrollElasticity(elasticity)` _macOS_ _Experimental_
 
-* `elasticity` string - Can be one of the following values: `automatic`, `none`, `allowed`. Default is `automatic`.
+* `elasticity` string - Can be `automatic`, `none`, `allowed`. Default is `automatic`.
 
 The scroll view’s horizontal scrolling elasticity mode.
 A scroll view can scroll its contents past its bounds to achieve an elastic effect. 
@@ -195,41 +205,62 @@ bounds only occurs if the document width is greater than the view width,
 or the vertical scroller is hidden and the horizontal scroller is visible.
 * `automatic` - Automatically determine whether to allow elasticity on this axis.
 * `none` - Disallow scrolling beyond document bounds on this axis.
-*`allowed` - Allow content to be scrolled past its bounds on this axis in an elastic fashion.
+* `allowed` - Allow content to be scrolled past its bounds on this axis in an elastic fashion.
 
-#### `scrollView.setVerticalScrollElasticity(elasticity)` _macOS_
+#### `scrollView.setVerticalScrollElasticity(elasticity)` _macOS_ _Experimental_
+
+* `elasticity` string - Can be `automatic`, `none`, `allowed`. Default is `automatic`.
 
 The scroll view’s vertical scrolling elasticity mode.
 
-#### `scrollView.getHorizontalScrollElasticity()` _macOS_
+#### `scrollView.getHorizontalScrollElasticity()` _macOS_ _Experimental_
 
 Returns `string` - The scroll view’s horizontal scrolling elasticity mode.
 
-#### `view.getVerticalScrollElasticity()` _macOS_
+#### `scrollView.getVerticalScrollElasticity()` _macOS_ _Experimental_
 
 Returns `string` - The scroll view’s vertical scrolling elasticity mode.
 
-#### `view.setScrollPosition(point)` _macOS_
+#### `view.setScrollPosition(point)` _Experimental_
 
 * `point` - The point in the `contentView` to scroll to.
 
 Scrolls the view’s closest ancestor `clipView` object so a point in the
 view lies at the origin of the clip view's bounds rectangle.
 
-#### `view.getScrollPosition()` _macOS_
+#### `view.getScrollPosition()` _Experimental_
 
 Returns `Point`
 
-#### `view.getMaximumScrollPosition()` _macOS_
+#### `view.getMaximumScrollPosition()` _Experimental_
 
 Returns `Point`
 
 #### Events
 
-#### `scroll-started`
-#### `scroll-ended`
-#### `bounds-did-change`
+#### Event: 'did-scroll' _Experimental_
 
+Returns:
+
+* `event` Event
+
+Emitted when the content view is being scrolled.
+
+#### Event: 'start-scroll' _Experimental_
+
+Returns:
+
+* `event` Event
+
+Emitted at the beginning of user-initiated scroll tracking.
+
+#### Event: 'end-scroll' _Experimental_
+
+Returns:
+
+* `event` Event
+
+Emitted at the end of scroll tracking.
 
 ## WrapperBrowserView
 
@@ -237,16 +268,16 @@ A `WrapperBrowserView` is the wrapper for `BrowserView`. It extends `BaseView`.
 
 Process: **Main**
 
-### `new WrapperBrowserView([options])`
+### `new WrapperBrowserView([options])` _Experimental_
 
 * `options` Object (optional)
   * `browserView` (optional)
 
-If `browserView` is not set then new `BrowserView` is created.
+If `browserView` is not set then a new `BrowserView` is created.
 
 ### Instance Properties
 
-#### `view.browserView`
+#### `view.browserView` _Experimental_
 
 A `BrowserView` object owned by this view.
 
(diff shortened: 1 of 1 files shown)


wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,178 @@
...
+class for all all different views. 
+
+Process: **Main**
+
+`BaseView` is an EventEmitter
+
+
+#### Static Methods
+
+* `BaseView.getAllViews()`
```
nornagon, line 65: What's the intended use case for this method?

daniel-koc: It returns all ScrollView's created - added for consistency (BaseView.getAllViews() -> ScrollView.getAllViews())

nornagon: I think this is kind of a misfeature in BaseView, I don't see the value in maintaining this list.

daniel-koc: We can remove this function from the proposal, no prob. The main goal for the API is to have the possibility to build hierarchy of views - ScrollView is the most important part.

nornagon: Cool, sounds good. Let's not add functions that have no clear purpose :)

daniel-koc: Removed - also did the small adjustment in the example :)

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,178 @@
...
+
+#### Events
+
+* `scroll-started`
+* `scroll-ended`
+
+
+### WrapperBrowserView
+
+A `WrapperBrowserView` is the wrapper for `BrowserView`. It extends [`BaseView`](base-view.md).
```
nornagon, line 164: Why is this wrapper necessary? It seems a bit awkward. Can we just use the BrowserView directly?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,178 @@
...
+* `setVerticalScrollElasticity(elasticity)` (_macOS_) - similar to `setHorizontalScrollElasticity`
+* `getHorizontalScrollElasticity()` (_macOS_)
+* `getVerticalScrollElasticity()` (_macOS_)
+* `scrollTo({ x, y })`
+
+
+#### Events
+
+* `scroll-started`
+* `scroll-ended`
```
nornagon, line 159: any others?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,178 @@
...
+
+Process: **Main**
+
+
+#### Static Methods
+
+
+* `ScrollView.fromId(id)`
+
+#### Instance Methods
```
nornagon, line 130: would you mind reformatting this section to match the format of our docs? It's a little hard to read as is.

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,178 @@
...
+
+Show a part of view with scrollbar.  The `ScrollView` can show an arbitrary content view inside it. It is used to make any View scrollable. When the content is larger than the `ScrollView`, scrollbars will be optionally showed. When the content view is smaller then the `ScrollView`, the content view will be resized to the size of the `ScrollView`. The scrollview supports keyboard UI and mousewheel. It extends `BaseView`.
+
+Process: **Main**
+
+
+#### Static Methods
+
+
+* `ScrollView.fromId(id)`
```
nornagon, line 128: Do we need a `fromId` method here? What's the use case for storing ID numbers as opposed to using a reference to the object itself?

wg-api/spec-documents/scroll-view.md
daniel-koc: I added WrapperBrowserView for clarity in the implementation. The thing is, on Mac, I defined the ElectronNativeView interface (inheriting from NSView), on the other hand, InspectableWebContentsViewMac creates ElectronInspectableWebContentsView.
I could get rid of WrapperBrowserView and add the addBrowserView / removeBrowserView functions to ContainerView instead.

wg-api/spec-documents/scroll-view.md
daniel-koc: I reformated the API description, hopefully it's better now.

wg-api/spec-documents/scroll-view.md
daniel-koc: Probably we will not need it; also we could ride of 'id' property too. I added BaseView.fromId() and `baseView.id` property for consistency with other Electron APIs.

nornagon: If there's not a clear purpose for the API, I think it's fine to remove it. The IDs on older APIs also don't serve a clear purpose IMO (perhaps they made more sense when `remote` was more popular?) so I don't think it's worth copying them for consistency here.

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+
+Process: **Main**
+
+`BaseView` is an EventEmitter
+
+### Static Methods
+
+#### `BaseView.getAllViews()`
+
+Returns the array of all created views.
```
nornagon, line 68: I thought this was dropped?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Instance Properties
+
+#### `view.id`
+
+A `Integer` property representing the unique ID of the view.
+
+#### `view.isContainer`
+
+A `boolean` property that determines whether this view is `ContainerView`.
```
nornagon, line 82: This is kind of weird, the JS way to do this would be to write `view instanceof ContainerView`.

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+
+#### `view.setBackgroundColor(color)`
+
+Change the background color of the view.
+
+#### `view.getParentView()`
+
+Returns `BaseView || null` - The parent view, otherwise returns `null`.
+
+#### `view.getParentWindow()`
```
nornagon, line 102: nit: I don't think `parent` is the best name here; that implies a direct relationship, but there are possibly several views in the hierarchy between the view and the window. Perhaps `getOwnerWindow` or just `getWindow`?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+
+Returns `BrowserWindow || null` - The window that the view belongs to, otherwise returns `null`.
+
+### Events
+
+#### Event: 'size-changed'
+
+Emitted when the view's size has been changed.
```
nornagon, line 110: What are the parameters to this event?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+### New instance methods
+
+#### `win.addChildView(view)`
+
+#### `win.removeChildView(view)`
+
+#### `win.getViews()`
+
+Returns `BaseView[]` - an array of all BaseViews that have been attached
+with `addChildView`.
```
nornagon, line 124: nit: perhaps `win.children`? Can this be a (read-only) property rather than a method?

nornagon, line 124: Does this include the WebContentsView?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+### Instance Methods
+
+#### `containerView.addChildView(view)`
+
+#### `containerView.removeChildView(view)`
+
+#### `containerView.getViews()`
+
+Returns `BaseView[]` - an array of all BaseViews that have been attached
+with `addChildView`.
```
nornagon, line 142: nit: I feel like these should just be `containerView.addChild(view)` / `containerView.removeChild(view)` / `containerView.children`. No need to repeat the word `view` over and over.

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+
+
+## ScrollView
+
+Show a part of view with scrollbar.
```
nornagon, line 147: > with scrollbar

does this depend on the OS / environment? e.g. on macOS when there's no external mouse plugged in, the default behavior is to hide the scrollbar, but this changes dynamically as external mice are added/removed.

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+
+## ScrollView
+
+Show a part of view with scrollbar. 
+The `ScrollView` can show an arbitrary content view inside it. It is used to make
+any View scrollable. When the content is larger than the `ScrollView`,
+scrollbars will be optionally showed. When the content view is smaller
+then the `ScrollView`, the content view will be resized to the size of the
+`ScrollView`.
+The scrollview supports keyboard UI and mousewheel.
```
nornagon, line 153: What keyboard UI? Is this customizable by an app?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+
+Process: **Main**
+
+### Static Methods
+
+#### `ScrollView.fromId(id)`
+
+Returns `ScrollView | null` - The scroll view with the given `id`.
+
+### Instance Methods
```
nornagon, line 164: Is there a constructor for this class?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+
+#### `scrollView.setContentView(contents)`
+
+Set the contents. The contents is the view that needs to scroll.
+
+#### `scrollView.getContentView()`
+
+Returns `BaseView` - The contents of the `scrollView`.
+
+#### `scrollView.setContentSize(size)`
```
nornagon, line 174: it seems a bit odd to me to set the size of the child view through a method on the parent. which way does this information flow? Does the content view choose its own size, or does the scroll view set the size of the content?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+
+Returns `string` - The scroll view’s horizontal scrolling elasticity mode.
+
+#### `view.getVerticalScrollElasticity()` _macOS_
+
+Returns `string` - The scroll view’s vertical scrolling elasticity mode.
+
+#### `scrollView.scrollTo(point)` _macOS_
+
+* `point` - The point in the `contentView` to scroll to.
```
nornagon, line 230: perhaps a way to get the current scroll position would be useful?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+More details of the use-case and challenges related to workarounds: [#32751](https://github.com/electron/electron/issues/32751)
+
+![](https://user-images.githubusercontent.com/2891206/150981006-84fe796e-60bd-43c0-952d-831d1bfd9998.gif)
+
+## Usage Example
+
+```js
+const { BrowserView, BrowserWindow, ContainerView, ScrollView, WrapperBrowserView } = require("electron");
+
+const window = new BrowserWindow({ width: 1000, height: 500 })
```
nornagon, line 28: Should this use `BaseWindow` instead of `BrowserWindow`?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,259 @@
...
+
+#### `win.removeChildView(view)`
+
+#### `win.getViews()`
+
+Returns `BaseView[]` - an array of all BaseViews that have been attached
+with `addChildView`.
+
+
+## ContainerView
```
nornagon, line 127: Should this just be called `View`? Does it do anything special?

wg-api/spec-documents/scroll-view.md
daniel-koc: OK, removing

wg-api/spec-documents/scroll-view.md
daniel-koc: I dropped for ScrollView, I can drop this one too.

wg-api/spec-documents/scroll-view.md
daniel-koc: I agree

wg-api/spec-documents/scroll-view.md
daniel-koc: That's the relationship; thus a view may be a child of either another view or a window. The best would be to have baseView.getParent() returning BaseView|BaseWindow.

wg-api/spec-documents/scroll-view.md
daniel-koc: I added for the possibility to react to a scrollview being scrolled by the user. The event is sent when the bounds of contentView's are changed. The new bounds is the parameter for this event.

wg-api/spec-documents/scroll-view.md
daniel-koc: Notice, that I didn't changed the API - so WrapperBrowserView (inheriting from BaseView) is still defined. So ContainerView doesn't have addBrowserView/removeBrowserView - to avoid the issue you mentioned above.
I discussed with Cheng - who proposed that we could move the implementation of NativeBrowserView::UpdateDraggableRegions implementation to InspectableWebContentsView - this way we could rid off  of using BrowserView (and remove WrapperBrowserView) - and instead define a WebContentsBrowserView API (inheriting from BaseView) with similar functionality to BrowserView.

wg-api/spec-documents/scroll-view.md
daniel-koc: I agree

wg-api/spec-documents/scroll-view.md
daniel-koc: No, it doesn't depend on the OS, removed this sentence.

wg-api/spec-documents/scroll-view.md
daniel-koc: I'm removing this sentence too. I think that I took it from views::ScrollView dscription, the description of API needs to be general.

wg-api/spec-documents/scroll-view.md
daniel-koc: Yes, adding to this spec

wg-api/spec-documents/scroll-view.md
nornagon: hmm, i disagree. a Window seems like a categorically different thing from a View, it would be weird to have one function to return either. the relationship is "ancestor" not "parent", no?

wg-api/spec-documents/scroll-view.md
daniel-koc: On Mac I defined this function as follows:
```
void NativeScrollView::SetContentSize(const gfx::Size& size) {
  auto* scroll = static_cast<ElectronNativeScrollView*>(GetNative());
  NSSize content_size = size.ToCGSize();
  [scroll setContentSize:content_size];
  [scroll.documentView setFrameSize:content_size];
}
```
That's the only way to set the content's size I think. Using setBounds for content of the scrollView causes the issue when scrollView's bounds are changed (scroll's content stops working).

wg-api/spec-documents/scroll-view.md
nornagon: So does it fall to the app to implement scrollbar show/hide on macOS? Do we expose the necessary APIs to detect when the scrollbar should be shown or hidden?

wg-api/spec-documents/scroll-view.md
daniel-koc: That's right. Actually I defined the functions: setScrollPosition, getScrollPosition, getMaximumScrollPosition. I'm adjusting this spec.

wg-api/spec-documents/scroll-view.md
daniel-koc: My current implementation works for both - BaseWindow and BrowserWindow. BrowserWindow has additional features - like triggering the before unload handlers etc. That's why I added to this spec baseWindow.addChild / baseWindow.removeChild (applicable for BrowserWindow too).

wg-api/spec-documents/scroll-view.md
daniel-koc: I agree. The think is that Electron has already `View` API - which assumes Chromium's views usage for all platforms. On the other hand we would need certain COCOA features, that's why I defined BaseView. Also - as I wrote before - we could merge ContainerView into BaseView.

wg-api/spec-documents/scroll-view.md
daniel-koc: I renamed getViews on getChildren - as I don't know what to do with BaseWindow.getChildren/getViews; I feel that it shouldn't be a property in such case (because what about BaseWindow.getBrowserViews?...)

wg-api/spec-documents/scroll-view.md
daniel-koc: Yes, scrollView.setHorizontalScrollBarMode and scrollView.setVerticalScrollBarMode

daniel-koc: Actually... I think that we should detecting properly, I'm not sure.

wg-api/spec-documents/scroll-view.md
daniel-koc: That's parent relationship :)

nornagon: I'm confused.

```
window
+-- view1
    +-- view2
```

`window` is not the parent of `view2`. I would argue that calling it the parent of `view1` is also inaccurate, because conflating the _window_ with the _view tree_ is not a good metaphor. The window has a single root view that has no parent, which is the top of the view tree.

calling `view2.getWindow()` should return `window`.

daniel-koc: Well, I agree, that should looks like that.
For the following scenario:
```
browserwindow1.addChild(view1)
containerView1.addChild(view2)
```
- view1.getParentView() will return null
- view2.getParentView() will return containerView1
I assumed, that
- view2.getParentWindow() will return null
- view1.getParentWindow() will return browserWindow1
in this way, it would be possible to distinguish to which object a view has been added.
We can replace getParentWindow by getWindow - then both view1.getWindow () and view2.getWindow () would return browserWindow1 - no prob.

wg-api/spec-documents/scroll-view.md
ckerr: The only reason for keeping them as separate classes is if we want to specifically disallow some View types from having children. e.g. do we want to disallow a ScrollView from having a child ScrollView.

There is some upside to that, but overall I think I'd prefer to merge the two classes so that the API has a smaller surface area.

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+
+Returns the `size` of the contents.
+
+#### `scrollView.setHorizontalScrollBarMode(mode)`
+
+* `mode` string - Can be one of the following values: `disabled`, `hidden-but-enabled`, `enabled`. Default is `enabled`.
+
+Controls how the horizontal scroll bar appears and functions.
+* `disable` - The scrollbar is hidden.
+* `hidden-but-enabled` - The scrollbar is hidden whether or not the contents are larger than the viewport, but the pane will respond to scroll events. This option is not available on MacOS.
```
ckerr, line 172: docs:

```suggestion
* `hidden-but-enabled` - The scrollbar is hidden whether or not the contents are larger than the viewport, but the pane will respond to scroll events. _Linux_  _Windows_
```

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+
+Set the size of the contents.
+
+#### `scrollView.getContentSize()`
+
+Returns the `size` of the contents.
+
+#### `scrollView.setHorizontalScrollBarMode(mode)`
+
+* `mode` string - Can be one of the following values: `disabled`, `hidden-but-enabled`, `enabled`. Default is `enabled`.
```
ckerr, line 168: docs:

```suggestion
* `mode` string - Can be `disabled`, `hidden-but-enabled`, `enabled`. Default is `enabled`.
```

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+
+Controls how the horizontal scroll bar appears and functions.
+* `disable` - The scrollbar is hidden.
```
ckerr, line 171: docs:

```suggestion
* `disabled` - The scrollbar is hidden.
```

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+* `hidden-but-enabled` - The scrollbar is hidden whether or not the contents are larger than the viewport, but the pane will respond to scroll events. This option is not available on MacOS.
+*`enabled` - The scrollbar will be visible if the contents are larger than the viewport.
+
+#### `scrollView.setVerticalScrollBarMode(mode)`
+
+Controls how the vertical scroll bar appears and functions.
+
+#### `getHorizontalScrollBarMode()`
```
ckerr, line 179: docs:

```suggestion
#### `scrollView.getHorizontalScrollBarMode()`
```

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+* `elasticity` string - Can be one of the following values: `automatic`, `none`, `allowed`. Default is `automatic`.
+
+The scroll view’s horizontal scrolling elasticity mode.
+A scroll view can scroll its contents past its bounds to achieve an elastic effect. 
+When set to `automatic`, scrolling the horizontal axis beyond its document
+bounds only occurs if the document width is greater than the view width,
+or the vertical scroller is hidden and the horizontal scroller is visible.
+* `automatic` - Automatically determine whether to allow elasticity on this axis.
+* `none` - Disallow scrolling beyond document bounds on this axis.
+*`allowed` - Allow content to be scrolled past its bounds on this axis in an elastic fashion.
```
ckerr, line 198: docs:

```suggestion
* `allowed` - Allow content to be scrolled past its bounds on this axis in an elastic fashion.
```

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+
+Returns `string` - horizontal scrollbar mode.
+
+#### `scrollView.getVerticalScrollBarMode()`
+
+Returns `string` - vertical scrollbar mode.
+
+#### `scrollView.setHorizontalScrollElasticity(elasticity)` _macOS_
+
+* `elasticity` string - Can be one of the following values: `automatic`, `none`, `allowed`. Default is `automatic`.
```
ckerr, line 189: docs:

```suggestion
* `elasticity` string - Can be `automatic`, `none`, `allowed`. Default is `automatic`.
```

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+
+#### `scrollView.setVerticalScrollElasticity(elasticity)` _macOS_
+
+The scroll view’s vertical scrolling elasticity mode.
+
+#### `scrollView.getHorizontalScrollElasticity()` _macOS_
+
+Returns `string` - The scroll view’s horizontal scrolling elasticity mode.
+
+#### `view.getVerticalScrollElasticity()` _macOS_
```
ckerr, line 208:
```suggestion
#### `scrollView.getVerticalScrollElasticity()` _macOS_
```

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+
+#### `view.getMaximumScrollPosition()` _macOS_
+
+Returns `Point`
+
+#### Events
+
+#### `scroll-started`
+#### `scroll-ended`
+#### `bounds-did-change`
```
ckerr, line 231: Do these have any parameters?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+A `WrapperBrowserView` is the wrapper for `BrowserView`. It extends `BaseView`.
+
+Process: **Main**
+
+### `new WrapperBrowserView([options])`
+
+* `options` Object (optional)
+  * `browserView` (optional)
+
+If `browserView` is not set then new `BrowserView` is created.
```
ckerr, line 245:
```suggestion
If `browserView` is not set then new a `BrowserView` is created.
```

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
+# Scroll View
+
+## Summary
+
+This spec proposes to add an experimental native ScrollView component to Electron.
```
ckerr, line 5: docs: all these APIs should be flagged with `_Experimental_`

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+The scroll view’s horizontal scrolling elasticity mode.
+A scroll view can scroll its contents past its bounds to achieve an elastic effect. 
+When set to `automatic`, scrolling the horizontal axis beyond its document
+bounds only occurs if the document width is greater than the view width,
+or the vertical scroller is hidden and the horizontal scroller is visible.
+* `automatic` - Automatically determine whether to allow elasticity on this axis.
+* `none` - Disallow scrolling beyond document bounds on this axis.
+*`allowed` - Allow content to be scrolled past its bounds on this axis in an elastic fashion.
+
+#### `scrollView.setVerticalScrollElasticity(elasticity)` _macOS_
```
ckerr, line 200: docs: need to copy the `Can be ...` section here too.

wg-api/spec-documents/scroll-view.md
ckerr: bikeshed: `enabled-but-hidden` feels slightly better than `hidden-but-enabled`

wg-api/spec-documents/scroll-view.md
ckerr: Is this need on Win / Linux, and do you have plans in that area?  This API seems pretty elaborate for single-platform use.

wg-api/spec-documents/scroll-view.md
ckerr: > I could get rid of WrapperBrowserView and add the addBrowserView / removeBrowserView functions to ContainerView instead.

I haven't thought through the implications so feel free to shoot this down, but could `addView()` just take `BaseView | BrowserView` instead and let the implementation call the right function  under the covers? If we have a new function named `addView()` then surely people are going to try passing BrowserViews into it

wg-api/spec-documents/scroll-view.md
ckerr: removed in b70f92aa91ae0991777332ac8af1ad34586ed0fc

wg-api/spec-documents/scroll-view.md
ckerr: removed in b70f92aa91ae0991777332ac8af1ad34586ed0fc

wg-api/spec-documents/scroll-view.md
ckerr: :+1: for instanceof; the `isFoo()` idiom doesn't scale as well if more View classes are added later

removed in b70f92aa91ae0991777332ac8af1ad34586ed0fc

wg-api/spec-documents/scroll-view.md
ckerr: removed in b70f92aa91ae0991777332ac8af1ad34586ed0fc

wg-api/spec-documents/scroll-view.md
ckerr: added in b70f92aa91ae0991777332ac8af1ad34586ed0fc

wg-api/spec-documents/scroll-view.md
ckerr: There is a lot of `_macOS_`-only API here and some of it, e.g. `view.setScrollPosition()`, seems important to having a well-behaved app. Is this functionality needed for Windows and Linux as well?

If I'm misunderstanding, e.g. if this is only relevant on macOS, maybe the docs could have sample code of what a well-behaved client that runs on all three would look like.

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+#### Event: 'size-changed'
+
+Emitted when the view's size has been changed.
+
+ 
+## BrowserWindow 
+
+### New instance methods
+
+#### `win.addChild(view)`
```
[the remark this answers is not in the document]
daniel-koc: > What happens if an already-parented child gets passed into this? Do we throw, return a pass/fail flag, etc
A child will be first detached from its current parent, and next added to a new one.
> What happens if when someone who's not developing in a strongly-typed environment passes a BrowserView into this? Does it error out, do we implicitly wrap it in a new WrapperBrowserView, etc.
For now exactly the same behaviour as in case of BaseWindow.addBrowserView: just does nothing (if the conversion fails)

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+
+#### `win.removeChild(view)`
```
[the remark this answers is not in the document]
daniel-koc: > What happens if passed in a view that is not a child -- is there any mechanism for the caller to know that this failed?

For now exactly the same behaviour as in case of BaseWindow.removeBrowserView - just does nothing

> Can this be used to remove any descendant view, or just direct children?

Just direct children

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,256 @@
...
+
+### `new ContainerView()`
+
+Creates the new container view.
+
+### Instance Methods
+
+#### `containerView.addChild(view)`
+
+#### `containerView.removeChild(view)`
```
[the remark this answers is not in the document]
daniel-koc: The same answers :)

wg-api/spec-documents/scroll-view.md
daniel-koc: That's right, this functionality should exist on all platforms. I have not yet defined how consistent behavior would be achieved, Chromium has View::ScrollRectToVisible defined. I will focus on that in the next step.

wg-api/spec-documents/scroll-view.md
daniel-koc: What about `containerView.getChildren`? Should it return only BaseViews?
Maybe a better idea would be to replace WrapperBrowserView by WebView - inheriting from BaseView - with an interface analogous to BrowserView?

wg-api/spec-documents/scroll-view.md
daniel-koc: fixed

wg-api/spec-documents/scroll-view.md
daniel-koc: fixed

wg-api/spec-documents/scroll-view.md
daniel-koc: updated the events description

wg-api/spec-documents/scroll-view.md
daniel-koc: fixed

wg-api/spec-documents/scroll-view.md
daniel-koc: fixed

wg-api/spec-documents/scroll-view.md
daniel-koc: fixed

wg-api/spec-documents/scroll-view.md
ckerr: moved from _macOS_ to _Experimental_ in 372905c240d9fc6953d2e7aefca0b1d6fcd2b795

wg-api/spec-documents/scroll-view.md
ckerr: Changed in 372905c240d9fc6953d2e7aefca0b1d6fcd2b795

wg-api/spec-documents/scroll-view.md
ckerr: fixed in 372905c240d9fc6953d2e7aefca0b1d6fcd2b795

wg-api/spec-documents/scroll-view.md
ckerr: fixed in 372905c240d9fc6953d2e7aefca0b1d6fcd2b795

wg-api/spec-documents/scroll-view.md
ckerr: fixed in 372905c240d9fc6953d2e7aefca0b1d6fcd2b795

wg-api/spec-documents/scroll-view.md
ckerr: fixed in 372905c240d9fc6953d2e7aefca0b1d6fcd2b795

wg-api/spec-documents/scroll-view.md
ckerr: fixed in 372905c240d9fc6953d2e7aefca0b1d6fcd2b795

wg-api/spec-documents/scroll-view.md
ckerr: fixed in 372905c240d9fc6953d2e7aefca0b1d6fcd2b795

wg-api/spec-documents/scroll-view.md
ckerr: fixed in 372905c240d9fc6953d2e7aefca0b1d6fcd2b795

wg-api/spec-documents/scroll-view.md
ckerr: fixed in 372905c240d9fc6953d2e7aefca0b1d6fcd2b795

zcbenz (approved):
API looks good to me.

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,281 @@
...
+
+
+## API Design
+
+In order to achieve the example above, we've made following changes to the API
+
+## BaseView
+
+A `BaseView` is a rectangle within the views View hierarchy. It is the base
+class for all all different views.
```
nornagon, line 56: What's the relationship between this and the existing `View` class?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,281 @@
...
+## Platforms
+* macOS 
+* Windows
+* Linux
+
+## Impetus
+
+It is currently not possible to add scrolling functionality around BrowserViews:
+
+* Capture mouse / touch events
```
nornagon, line 16: I don't see any new events for capturing e.g. a click on a view. Is that out of scope for this?

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,281 @@
...
+
+Emitted at the beginning of user-initiated scroll tracking.
+
+#### Event: 'end-scroll' _Experimental_
+
+Returns:
+
+* `event` Event
+
+Emitted at the end of scroll tracking.
```
erickzhao, line 260: * `did-scroll` and `end-scroll` are pretty similar in naming. Maybe something like `will-scroll` (before), `scroll` (during), and `did-scroll` (after) would work better?

* More detail for each event similar to https://www.electronjs.org/docs/latest/api/web-contents#event-will-navigate would be helpful. For instance, will the scroll be able to be cancelled via `event.preventDefault`?

wg-api/spec-documents/scroll-view.md
daniel-koc: I couldn't make BaseView and View compatible - because COCOA used for BaseView (on Mac). So there is no relationship.

wg-api/spec-documents/scroll-view.md
daniel-koc: It is out of scopefor this spec; we extended the implementation for such functionalities.

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,281 @@
...
+* `options` Object (optional)
+  * `webPreferences` (optional)
+
+### Instance Properties
+
+#### `view.webContents` _Experimental_
+
+A `WebContents` object owned by this view.
+
+## Rollout Plan
```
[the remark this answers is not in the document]
daniel-koc: > Can we want to hide this behind a command line switch at first? This would make it more explicit for users that this API is subject to change.
Would you like to hide  all APIs (BaseView, ScrollView, WebBrowserView) behind a command line switch? I think that would be possible.
> • Document existing View APIs that landed from @zcbenz
> • Next-gen view API replacing legacy BrowserView 
Looks reasonable for me.
> ◦ For stability: how do we maintain backward-compatibility for the deprecated BrowserView? Can we make it a JS API layer around the new WebBrowserView?
We could do it in that way. The previously proposed WrapperBrowserView would provide somehow  backward-compatibility for the BrowserView, we could add 'browserView' to the options of WebBrowserView. We would not be able add BrowserView to the hierarchy of BaseView directly.

wg-api/spec-documents/scroll-view.md
nornagon: Wait, what cocoa-specific features are in BaseView? I would think those could be integrated into View. We should NOT have two _different_ View base classes.

daniel-koc: At https://github.com/electron/electron/issues/32751#issuecomment-1034811577 we were discussing with @zcbenz this topic, he agreed for the possibility  of use NSView on macOS and views::View on others.
There are several reasons why I have designed the current implementation in this direction:
- vibrancy effect on Mac for views,
- animations for setting the bounds, scaling etc. easily implementable in Cocoa,
- some other visual properties, like shapes for NSView;
Most of these features can also be achieved with Chromium views, but probably not all.

nornagon: I don't think we want to expose any NSView stuff directly. If we need extra features from views that views doesn't already support, I think it could make sense to patch those things in or otherwise extend views to support them. But I think it is a bad idea to have one API that backs onto views on win/linux and onto NSView/Cocoa on mac. Chrome used to do this and moved away from it.

Views [supports animations](https://source.chromium.org/chromium/chromium/src/+/main:ui/views/animation/); I'm not sure what you mean by "shaping" but `View::SetClipPath` is a thing. Vibrancy isn't supported by views, but it seems like something we could build on top of views if needed. I don't think it calls for a whole separate implementation.

daniel-koc: There is a lot of bugs with the Chromium's ScrollView, for example UiCompositorScrollWithLayers (which provides the dezired performance for WebViews) is implemented only on MacOs. But currently we're working on it, we already have a ready solution for Windows, we also implement other features (to be consistent with Mac) - such as elasticity. So we agree that in this situation the best solution is to use Chromium views on all platforms - and implement the missing features.

daniel-koc: Fixed in 8418e0413a4cccc2258add477052713d647a9300

wg-api/spec-documents/scroll-view.md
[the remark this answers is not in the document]
daniel-koc: Added the few points in 8418e0413a4cccc2258add477052713d647a9300

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,281 @@
...
+scrollContent.addChild(webBrowserView2)
+
+scroll.setContentView(scrollContent);
+win.addChild(scroll);
+```
+
+
+## API Design
+
+In order to achieve the example above, we've made following changes to the API
```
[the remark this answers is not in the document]
daniel-koc: Added in 8f0179291c9f2a51ec2b529b9b56872c521eb704

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,281 @@
...
+
+#### `scrollView.getVerticalScrollElasticity()` _macOS_ _Experimental_
+
+Returns `string` - The scroll view’s vertical scrolling elasticity mode.
+
+#### `view.setScrollPosition(point)` _Experimental_
+
+* `point` - The point in the `contentView` to scroll to.
+
+Scrolls the view’s closest ancestor `clipView` object so a point in the
```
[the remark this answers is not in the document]
daniel-koc: An internal object that clips a document view to a scroll view's frame. (taken from Apple documentation)

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,281 @@
...
+
+Creates the new scroll view.
+
+### Instance Methods
+
+#### `scrollView.setContentView(contents)` _Experimental_
+
+Set the contents. The contents is the view that needs to scroll.
+
+#### `scrollView.getContentView()` _Experimental_
```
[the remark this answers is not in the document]
daniel-koc: This is the view the scroll view scrolls within its content view.

wg-api/spec-documents/scroll-view.md
daniel-koc: This is for revisit in future - as currently taken from Apple documentation, it will need some adjustment when implementing by Chromium views.

wg-api/spec-documents/scroll-view.md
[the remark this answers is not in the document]
daniel-koc: Fixed in e7a24bd167633310339960ae36263c88f3657388

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,281 @@
...
+* `point` - The point in the `contentView` to scroll to.
+
+Scrolls the view’s closest ancestor `clipView` object so a point in the
+view lies at the origin of the clip view's bounds rectangle.
+
+#### `view.getScrollPosition()` _Experimental_
+
+Returns `Point`
+
+#### `view.getMaximumScrollPosition()` _Experimental_
```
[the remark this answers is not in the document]
daniel-koc: Fixed in e7a24bd167633310339960ae36263c88f3657388

wg-api/spec-documents/scroll-view.md
daniel-koc: Restored the name `8418e0413a4cccc2258add477052713d647a9300` - currently defined in `View`.

wg-api/spec-documents/scroll-view.md
```diff
@@ -0,0 +1,265 @@
...
+Returns:
+
+* `event` Event
+
+Emitted at the end of scroll tracking.
+
+## Rollout Plan
+
+- add the documentation to [Views API (Part 1)](views-api-1-content-view.md)
+- Move the implementation from `BrowserView` to `InspectableWebContentsView` (in particular the implementation of draggable regions)
```
erickzhao, line 263: I'd like to learn more about `InspectableWebContentsView`, haven't seen it mentioned elsewhere in the RFC 😮

wg-api/spec-documents/scroll-view.md
erickzhao: Do you mean the naming convention is from Apple? Is this specific to Cocoa? Sorry if I'm misunderstanding!

daniel-koc: Not really; slighty adjusted in 4b573a32b643c8f224c418b9aa83a12c65b7e210

wg-api/spec-documents/scroll-view.md
Sami-ab: []().```RAD.E¥E.S5¥```.

ckerr (approved):

PR #483, opened 2022-06-23
chore: fix links in issue triage playbook

dsanders11:
A handful of links were broken, mostly due to crossing in-flight with #480.

Also cleaned up the link text in a few cases to make sure it matches to response template names.

We'll get a docs link linter on this repo at some point. 😅


VerteDinde:
Thanks for fixing these! 🙇‍♀️

PR #484, opened 2022-06-27
docs: Electron C APIs

zcbenz:
This is a proposal for adding C APIs that can be used to interact with internal Electron and Chromium in native modules.

### DIFF
--- a/wg-api/spec-documents/electron-c-apis.md
+++ b/wg-api/spec-documents/electron-c-apis.md
@@ -16,7 +16,7 @@ There are 2 reasons:
 
 The majority of Electron’s JavaScript APIs are about accessing native resources of the operating system; some of them could not be implemented as native modules because they need to interact with Electron and Chromium’s resources.
 
-For example, the `TouchBar` and `inAppPurchase` APIs must live as built-in APIs because the JavaScript bindings must interact with internal `NSApp` and `NSWindow`, and there is no non-intrusive way doing that outside Electron.
+For example, the `TouchBar` and `inAppPurchase` APIs must live as built-in APIs because the JavaScript bindings must interact with internal `ElectronApplicationDelegate` and `ElectronNSWindowDelegate`, and there is no non-intrusive way doing that outside Electron.
 
 With Electron C APIs we can make it possible to have those APIs as native modules. This can make Electron less bloating since we don't have to make every new API a built-in API , and for users who want to add a large set of new APIs they can now only request a few minimal C APIs in Electron and then implement anything they want in native modules.
 
@@ -28,11 +28,7 @@ Some people may already know that the official releases of VS Code and Teams use
 
 __Note that this spec does not define any speicific API and the C APIs used in this spec are only for demonstration, the actual APIs should be discussed in follwoup PRs.__
 
-The Electron C APIs should be very cheap: they only take minimal disk space, usually less than 40 bytes per API since they are just thin wrappers; and they have no runtime cost.
-
-The Electron C APIs should need no API desgin: they are mostly direct C wrappers of C++ interface.
-
-The Electron C APIs should be easy to pass API review: if there is an actual usage we should be able to approve a C API quickly, since they are usually cheap and exposing an existing API.
+The Electron C APIs should be very cheap: they only take minimal disk space, usually less than 100 bytes per API since they are just thin wrappers; and they have no runtime cost.
 
 To use the Electron C APIs in native modules, users only need to include the related headers:
 
--- a/wg-api/spec-documents/electron-c-apis.md
+++ b/wg-api/spec-documents/electron-c-apis.md
@@ -1,0 +1,161 @@
+# Electron C APIs
+
+## Summary
+
+Similar to the [Node-API (also known as N-API) of Node.js](https://nodejs.org/api/n-api.html), the Electron C APIs is a set of C APIs providing access to Electron and Chromium’s resources, which can be used in Node.js native modules.
+
+## Platforms
+
+All
+
+## Impetus
+
+There are 2 reasons:
+
+### Easier to add new APIs
+
+The majority of Electron’s JavaScript APIs are about accessing native resources of the operating system; some of them could not be implemented as native modules because they need to interact with Electron and Chromium’s resources.
+
+For example, the `TouchBar` and `inAppPurchase` APIs must live as built-in APIs because the JavaScript bindings must interact with internal `NSApp` and `NSWindow`, and there is no non-intrusive way doing that outside Electron.
+
+With Electron C APIs we can make it possible to have those APIs as native modules. This can make Electron less bloating since we don't have to make every new API a built-in API , and for users who want to add a large set of new APIs they can now only request a few minimal C APIs in Electron and then implement anything they want in native modules.
+
+### Reduce Electron forks
+
+Some people may already know that the official releases of VS Code and Teams use custom builds instead of the offical one, this is because they have some modifications that can not be upstreamed (for example some hacks to the `<video>` tag). By exposing certain internal Chromium APIs as Electron C APIs, we can make those modifications live as native modules instead of having a custom build.
+
+## API Design
+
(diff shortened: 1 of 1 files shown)


wg-api/spec-documents/electron-c-apis.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Impetus
+
+There are 2 reasons:
+
+### Easier to add new APIs
+
+The majority of Electron’s JavaScript APIs are about accessing native resources of the operating system; some of them could not be implemented as native modules because they need to interact with Electron and Chromium’s resources.
+
+For example, the `TouchBar` and `inAppPurchase` APIs must live as built-in APIs because the JavaScript bindings must interact with internal `NSApp` and `NSWindow`, and there is no non-intrusive way doing that outside Electron.
```
nornagon, line 19: For `NSWindow`, it's possible to get the pointer to the native window out of the BrowserWindow handle, right? https://www.electronjs.org/docs/latest/api/browser-window#wingetnativewindowhandle

wg-api/spec-documents/electron-c-apis.md
```diff
@@ -0,0 +1,161 @@
...
+
+Some people may already know that the official releases of VS Code and Teams use custom builds instead of the offical one, this is because they have some modifications that can not be upstreamed (for example some hacks to the `<video>` tag). By exposing certain internal Chromium APIs as Electron C APIs, we can make those modifications live as native modules instead of having a custom build.
+
+## API Design
+
+__Note that this spec does not define any speicific API and the C APIs used in this spec are only for demonstration, the actual APIs should be discussed in follwoup PRs.__
+
+The Electron C APIs should be very cheap: they only take minimal disk space, usually less than 40 bytes per API since they are just thin wrappers; and they have no runtime cost.
+
+The Electron C APIs should need no API desgin: they are mostly direct C wrappers of C++ interface.
```
nornagon, line 33: This seems questionable; the underlying APIs are subject to change, but wrappers will need to be somewhat stable (though not necessarily ABI-stable, as we expect to rebuild native modules for each version of Electron), so there may be some careful decisions to be made.

wg-api/spec-documents/electron-c-apis.md
```diff
@@ -0,0 +1,161 @@
...
+
+// Get the native window handle of Electron window.
+NSWindow* window = electron_base_window_get_native_handle(object);
+[window setToolbar:toolbar];
+
+// Create native views.
+electron_view_t view = electron_view_create();
+electron_button_t button = electron_button_create("Title");
+electron_view_add_child_view(view, button);
+electron_window_set_content_view(window, view);
```
nornagon, line 56: What schema do you imagine for anything that needs to subscribe to events?

These kinds of objects currently have a V8-managed lifetime. How should the C APIs expose GC-related primitives? Does each C handle map to a V8 handle under the hood, so that the objects remain referenced and don't get garbage collected? Or do we provide some other system for C APIs to interact with GC (e.g. refcounting)?

nornagon (changes requested):
This seems like an interesting direction!

Taking it a bit further, could we imagine building all the JS APIs on top of these C APIs, instead of directly?

Further, could we delete all our C++ code and just depend on CEF? :)

deermichel:
THIS IS AWESOME 🤩  !!! - i know it's just a proposal for now, but do you have any idea of a timeline @zcbenz ? keep me posted, i might join the effort

itsananderson:
This is a great idea, and I can immediately think of a bunch of custom Electron patches we could clean up if APIs like this were in place.

Regarding the following comment from @nornagon:
> we expect to rebuild native modules for each version of Electron

Our team has actually finally reached a point where we don't need to rebuild our native modules for each version of Electron, since we leverage N-API. Would it be possible for these Electron C APIs to be exposed in a similarly ABI compatible way so that native modules wouldn't have to be rebuilt, or is there some non-obvious reason it would be infeasible?

samuelmaddock (commented):
Fascinating idea, one which is going to have me rethink future API proposals. Would this approach support usage in both the main and renderer process?

nornagon:
How will C APIs work in the renderer? The sandbox can't load shared libraries. Would C APIs in the renderer be restricted to unsandboxed renderers?

nornagon:
@zcbenz it's possible to add specific exceptions to the sandbox, though i'm not sure how easy it would be to do so dynamically—I believe all such exceptions must be added prior to the sandbox being enabled for a particular process, so specific modules would have to be carved out ahead-of-time.

that said, I don't think there's a huge need for C modules in sandboxed renderers; they ought to be doing all their comms with the outside world over IPC to the main process anyway.

MarshallOfSound:
> that said, I don't think there's a huge need for C modules in sandboxed renderers; they ought to be doing all their comms with the outside world over IPC to the main process anyway.

I'm in the other camp here, this proposal in general was getting a `-0` from me mentally until I remembered a use case, namely custom media stream sources from native code (Similar to how Discords Electron fork operates).  The way to do that performantly is to allow a native module / external library to link into Electron and provide frame data directly into the renderer.  This means we'd need to solve the sandboxed-renderer issue 🤔 

Ref: https://github.com/discord/electron/blob/main/discord/discord_video_source.cc#L253-L265

clavin (approved):
I'm in favor of this proposal. I think the idea of cutting down on forks is a great driving motivation for this work and I want to see how far we can take this idea.

nornagon:
@MarshallOfSound I feel like that sort of use case ought to be addressed by transferring either HANDLEs or shared memory from the main process via mojo. e.g. [mojo::PlatformHandle](https://source.chromium.org/chromium/chromium/src/+/main:mojo/public/cpp/platform/platform_handle.h;l=28-42;drc=8adc9cb73b65f12d089dabbad06a9fb7446e478f) or [mojo::SharedBufferHandle](https://source.chromium.org/chromium/chromium/src/+/main:mojo/public/cpp/system/buffer.h;l=43-45;drc=8adc9cb73b65f12d089dabbad06a9fb7446e478f).

wg-api/spec-documents/electron-c-apis.md
```diff
@@ -0,0 +1,157 @@
+# Electron C APIs
+
+## Summary
+
+Similar to the [Node-API (also known as N-API) of Node.js](https://nodejs.org/api/n-api.html), the Electron C APIs is a set of C APIs providing access to Electron and Chromium’s resources, which can be used in Node.js native modules.
```
bnb, line 5: > Similar to the [Node-API of Node.js](https://nodejs.org/api/n-api.html), the Electron C APIs is a set of C APIs providing access to Electron and Chromium’s resources, which can be used in Node.js native modules.

This is technically now Node-API, not what's defined here.

wg-api/spec-documents/electron-c-apis.md
```diff
@@ -0,0 +1,157 @@
...
+
+There are 2 reasons:
+
+### Easier to add new APIs
+
+The majority of Electron’s JavaScript APIs are about accessing native resources of the operating system; some of them could not be implemented as native modules because they need to interact with Electron and Chromium’s resources.
+
+For example, the `TouchBar` and `inAppPurchase` APIs must live as built-in APIs because the JavaScript bindings must interact with internal `ElectronApplicationDelegate` and `ElectronNSWindowDelegate`, and there is no non-intrusive way doing that outside Electron.
+
+With Electron C APIs we can make it possible to have those APIs as native modules. This can make Electron less bloating since we don't have to make every new API a built-in API , and for users who want to add a large set of new APIs they can now only request a few minimal C APIs in Electron and then implement anything they want in native modules.
```
bnb, line 21:
```suggestion
With Electron C APIs we can make it possible to have those APIs as native modules. This can make Electron less bloating since we don't have to make every new API a built-in API, and for users who want to add a large set of new APIs they can now only request a few minimal C APIs in Electron and then implement anything they want in native modules.
```

nornagon (approved):

nornagon:
seems legit

MarshallOfSound (approved):

PR #486, opened 2022-07-18
docs: add renderer scripts spec

samuelmaddock:
Proposal to add a new scripting context to renderers to enable safe interactions with non-render frame contexts.

Better reader view: https://github.com/electron/governance/blob/53372bfb3fecc2daf4e0a2bc1371342157056444/wg-api/spec-documents/renderer-scripts.md


samuelmaddock:
Will open again as an RFC.

PR #508, opened 2022-09-22
docs: add wg-infra readme

MarshallOfSound:
As discussed at Summit, this lays the groundwork for a new place for docs, policies and a new organizational structure around the things this working group will own that were previously "unowned"

### DIFF
--- a/wg-infra/README.md
+++ b/wg-infra/README.md
@@ -1,0 +1,42 @@
+# Infrastructure WG
+
+Oversees all infrastructure, bots, automation, services and systems that support the Electron maintainers group.
+
+## Membership
+
+| Avatar | Name | Role | Time Zone |
+| -------------------------------------------|----------------------|----------------------------| -------- |
+| <img src="https://github.com/marshallofsound.png" width=100 alt="@marshallofsound">  | Samuel Attard [@MarshallOfSound](https://github.com/marshallofsound) | Member | PT (Vancouver) |
+| <img src="https://github.com/codebytere.png" width=100 alt="@codebytere">  | Shelley Vohr [@codebytere](https://github.com/codebytere) | Member | CET (Berlin) |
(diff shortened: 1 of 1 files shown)


wg-infra/README.md
```diff
@@ -0,0 +1,43 @@
...
+  * GitHub
+  * Grafana
+  * Goma
+  * CircleCI
+  * Appveyor
+  * Sentry
+  * Slack
+  * 1Password
+  * NPM
+* Electron Bots
```
codebytere, line 28: We should be more specific about which bots

codebytere (approved):

wg-infra/README.md
MarshallOfSound: All the bots? Should I list them all 😅 Or maybe it's safer to say we own the infrastructure the bots run on, and deployments, but not the code itself?

codebytere: That seems slightly better - let's do that

wg-infra/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+In order to join the Infrastructure Working Group, an aspiring member must:
+
+1. Show a continued interest in the systems this group is responsible for
+2. Be a long-trusted member of the maintainers community
+3. TBD, this group is still forming primarily in an organizational capacity, when the initial storm is over this list will be updated with a more accurate set of requirements
+
+## WG Removal Policy
+
+If a sitting member of the WG has not been active in a meaningful way for at least one month, the WG may vote to remove them from its set of sitting members.
```
erickzhao, line 41: We're a pretty high-trust group so I think this has a pretty low chance of happening, but we should probably formalize what determines the success or failure of said vote.

dsanders11:
Would the `build-tools` repo be owned by this WG? I think it's one of the things that is currently "unowned" by any WG. IIRC @codebytere and I had a hallway conversation about which WG should own `build-tools` and it might have been this one.

MarshallOfSound:
@dsanders11 added it (and a few other repos) to the list

PR #520, opened 2022-11-04
refactor: charter updates - accuracy

clavin:
This PR sets out to update the charter with the goal of improving its **Accuracy** in regards to how governance has been working in practice. Major changes are as follows:

* The Rotating Chair Model was refactored out (but not changed significantly) to allow working groups to choose their own model of accomplishing work.
  * This is supported by working groups where the purpose of a chair wasn't providing value.
  * One modification made to the Rotating Chair Model was the ability for a working group to decide its own term length for the chair.
* The concept of a *Delegate* was added to fulfill the governance-related roles that were previously expected by the chair, which not every working group may now have.
* Minor fix-ups to links and comments were made.

Let me know if anything else should be changed or if any of these changes don't fit with our model of governance! I would be more than happy to receive everyone's feedback here. 😊

### DIFF
--- a/charter/rotating-chair-model.md
+++ b/charter/rotating-chair-model.md
@@ -1,0 +1,57 @@
+# Rotating Chair Model
+
+The Rotating Chair Model is a model of a working group that uses a chosen leader to push forward the working group's goals. The chosen leader is regularly replaced with a new one, creating a "rotation" of leadership among the working group's members. This rotation promotes knowledge sharing and distribution of work duties among members.
+
+## Working Group Responsibilities
+
+A Working Group that uses the Rotating Chair Model has an additional core responsibility:
+* They shall [select](#Chair-Terms-and-Selection) a rotating chair to [represent](#Chair-Responsibilities) the group.
+
+## Chair Terms and Selection
+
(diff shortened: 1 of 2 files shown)


charter/rotating-chair-model.md
```diff
@@ -0,0 +1,57 @@
...
+
+In addition to these responsibilities, a chair may have additional responsibilities specific to their Working Group.
+
+## Is the Rotating Chair Model Right For My Working Group?
+
+(This section is non-normative.)
+
+The following advice and considerations can help a new Working Group decide whether the Rotating Chair Model is a good fit:
+
+*
```
nornagon, line 57: ?

nornagon (commented):
Broadly LGTM! I think there are changes we should make to how chairs/delegates work, but this seems better than what we currently have.

charter/rotating-chair-model.md
clavin: This part was still WIP! Feel free to look at the new text here.

nornagon (approved):

PlaineKevin (approved):
LGTM as well!

PR #521, opened 2022-11-14
Prefer using milliseconds for time measurement

samuelmaddock:
https://github.com/electron/electron/pull/35987 introduces an API which accepts a time measurement. We should set a standard for all APIs going forward.

Using the [web platform's design principles](https://w3ctag.github.io/design-principles/#milliseconds) as a guide, I'm proposing we use milliseconds.

### DIFF
--- a/wg-api/best-practices.md
+++ b/wg-api/best-practices.md
@@ -111,6 +111,13 @@ If you need to express a state which is a combination of properties, which might
 
 See https://w3ctag.github.io/design-principles/#string-constants.
 
+### Use milliseconds for time measurement
+If you are designing an API that accepts a time measurement, express the time measurement in milliseconds.
+
+Even if seconds (or some other time unit) are more natural in the domain of an API, sticking with milliseconds ensures that APIs are interoperable with one another. This means that authors don’t need to convert values used in one API to be used in another API, or keep track of which time unit is needed where.
+
+https://w3ctag.github.io/design-principles/#milliseconds
+
 ## Classes
 
 ### Use a class's name as the module name


erickzhao (approved):

jkleinsc (approved):

nornagon (approved):
FWIW we have several APIs that use/return stuff in seconds:

* [`DownloadItem.getStartTime`](https://www.electronjs.org/docs/latest/api/download-item#downloaditemgetstarttime) returns seconds since UNIX epoch.
* [`UploadFile.modificationTime`](https://www.electronjs.org/docs/latest/api/structures/upload-file) is in seconds since UNIX epoch.
* [`session.createInterruptedDownload`](https://www.electronjs.org/docs/latest/api/session#sescreateinterrupteddownloadoptions) takes `startTime` as seconds since UNIX epoch.
* [`cookies.set`](https://www.electronjs.org/docs/latest/api/cookies#cookiessetdetails) takes an `expirationDate` in seconds since UNIX epoch.
* [`Cookie.expirationDate`](https://www.electronjs.org/docs/latest/api/structures/cookie) is in seconds since UNIX epoch.
* [`powerMonitor.getSystemIdleState`](https://www.electronjs.org/docs/latest/api/power-monitor#powermonitorgetsystemidlestateidlethreshold) takes seconds for `idleThreshold`.
* [`powerMonitor.getSystemIdleTime`](https://www.electronjs.org/docs/latest/api/power-monitor#powermonitorgetsystemidletime) returns a duration in seconds.
* [`Certificate.validStart`/`validExpiry`](https://www.electronjs.org/docs/latest/api/structures/certificate) are seconds since UNIX epoch.

vs. the APIs we have that are in milliseconds:

* [`process.getCreationTime()`](https://www.electronjs.org/docs/latest/api/process#processgetcreationtime) is in milliseconds since UNIX epoch.
* [`PaymentDiscount.timestamp`](https://www.electronjs.org/docs/latest/api/structures/payment-discount) is in milliseconds since UNIX epoch.
* [`ProcessMetric.creationTime`](https://www.electronjs.org/docs/latest/api/structures/process-metric) is in milliseconds since UNIX epoch.

I don't mind standardizing on milliseconds for future APIs, but we should put a warning note on the existing APIs that use seconds to call out that they differ from the standard. It's a little awkward that we have many more APIs that use seconds than that use milliseconds, but _c'est la vie_ I suppose.

PR #533, opened 2023-02-13
feat: add spec for login item revamp

codebytere:
Initial spec for a revamp of `app.{set|get}LoginItemSettings(settings)`.


wg-api/spec-documents/login-item-api.md
```diff
@@ -0,0 +1,80 @@
...
+## Platforms
+
+**macOS**
+
+## Impetus
+
+At present, Electron uses `LSSharedFileList` APIs to register applications as login items. These APIs have been deprecated since 10.11, and have not worked for several years on sandboxed Mac App Store builds. Chromium is also slowly [removing usage upstream](https://chromium-review.googlesource.com/c/chromium/src/+/3997795), requiring us to float increasingly deprecated code. As of Ventura, several key aspects of login item settings, 
+namely `openAsHidden`, also no longer work.
+
+In macOS 13, Apple introduced [`SMAppService`](https://developer.apple.com/documentation/servicemanagement/smappservice?language=objc) to take the place of the APIs they phased out. This service is used to register and control `LoginItems`, `LaunchAgents`, and `LaunchDaemons` as helper executables for apps.
```
nornagon, line 18: Is this API restricted to working on macOS 13+? We still support macOS 12, so we'll have to support the older APIs until we eventually drop support for 12, then, right?

wg-api/spec-documents/login-item-api.md
```diff
@@ -0,0 +1,80 @@
...
+
+## API Design
+
+The new methods will adhere to the following semantics:
```
nornagon, line 22: It looks like these have the same name & a (nearly) compatible type signature with the existing methods, is that right?

bitdisaster:
This change should support backwards compatibility/migration considering falling scenarios:
When launch on login was enabled with old version:
1. New version should report the setting via old API.
2. Disabling in new version should disable via old API
3. Enabling again should disable in the old API and enable in new API

wg-api/spec-documents/login-item-api.md
codebytere: Yes - we would need to support them until Chromium drops support for 12. I can add that to the document to make it more explicit/clear!

wg-api/spec-documents/login-item-api.md
codebytere: Yes - just property differences!

codebytere:
@bitdisaster oh good points - I can update the document to reflect what the scenarios would be in those cases!

erickzhao (commented):
I don't think we _need_ this at this stage, but having a migration guide for users of the removed APIs would be good.

Maybe it could be something as simple as a guidance on telling them to implement "open as hidden" as an app-level setting.

wg-api/spec-documents/login-item-api.md
codebytere: done.

erickzhao (approved):

codebytere:
https://github.com/electron/electron/pull/37244 has now been merged.

PR #534, opened 2023-02-16
docs: add a "Release Captain" guide

VerteDinde:
This PR adds a guide for being a Release Captain in the Releases WG.


wg-releases/release-captain-overview.md
```diff
@@ -0,0 +1,65 @@
...
+## Rotation
+
+Each Release Captain Rotation lasts ~8 weeks. All current members of the Releases Working Group have been added to the Release Captain rotation.
+
+If you're not able to be the Release Captain during a specific period, or need others to cover your shift during specific weeks, post in the Releases WG Slack channel to find coverage.
+
+## Responsibilites
+
+### Ongoing (Week 1-8)
+* Review the current pre-release branch [project board] for new items
```
dsanders11, line 22: nit: could all of these links be changed to be of the style `[project board][]` and `[meeting agenda][]`?

Would match what we enforce on `e/e`, and makes linting easier.

dsanders11 (approved):

codebytere (changes requested):
As written, I think it's reasonable to say that this role comprises the vast bulk of duties previously covered by the WG Chair - as such, I think we need to decide whether to proceed with a formal WG chairship for this WG.

If the answer is yes, we need to update the [Chair Responsibilites](https://github.com/electron/governance/blob/868a501d0b746f233dc6f8d46a9f29f766c18837/wg-releases/README.md#L35) to reflect changes as a result of this role as well as decide what roles are specific to the Chair. If we don't think we should continue with a Chair, we should ensure duties listed there are enumerated in this PR. We should also link here from the main README.

jkleinsc (approved):

codebytere (approved):

PR #563, opened 2023-11-15
chore: add emeritus members to working groups

VerteDinde:
This PR follows up on several off boarding items discussed at the October Boston Electron Summit:

- Adds a status for emeritus, or former members of working groups.
- Adds a description for what qualifies a member for emeritus status
- Explicitly defines a process for removing inactive maintainers from working groups.

This PR does not define what an inactive member means - that definition should be determined individually by each working group, and added to individual working group READMEs.

_Note: The members listed here as emeritus were pulled from git and Slack history, and may not be a complete list. If you feel you've been added in error, or are not on the list and should be, please reach out or comment here. 🙏_

### DIFF
--- a/charter/README.md
+++ b/charter/README.md
@@ -143,4 +143,4 @@ For the safety and security of the project, all maintainers who have been offboa
 
 ### Emeritus Status
 
-When a former maintainer is offboarded, they will be automatically moved to Emeritus status within the governance repo. There is no minimum required period of servicing as a working group member for this status - any working group member who served for any length of time, but is no longer part of the working group, is eligible.
+When a former maintainer is offboarded, they will be automatically moved to Emeritus status within the governance repo. There is no minimum required period of serving as a working group member for this status - any working group member who served for any length of time, but is no longer part of the working group, is eligible.
--- a/charter/README.md
+++ b/charter/README.md
@@ -130,3 +130,17 @@ If a delegate Working Group is failing to meet its responsibilities, the AWG may
 
 > **Authors note:**
 > The current Electorate/AWG mix places almost all the power in the hands of a few large corporations at the outset. While not ideal to many, this was chosen because it represents the present status-quo, as of writing. The choice to delay defining an Electorate was made to _ensure_ it was not rushed, and that we have time to balance project stakeholders large and small.
+
+## Onboarding / Offboarding
+
+### Offboarding
+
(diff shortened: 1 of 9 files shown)


charter/README.md
```diff
@@ -130,3 +130,17 @@ If a delegate Working Group is failing to meet its responsibilities, the AWG may
...
+
+If a maintainer no longer wishes to be a part of a working group, they should let the working group know via Slack or by submitting a PR removing themselves from the working group in governance. If a maintainer becomes inactive, the working group may choose to vote the member out via their own process - this process will be defined by each individual working group.
+
+For the safety and security of the project, all maintainers who have been offboarded from all working groups will be removed from GitHub member status and GSuite. They may also be moved to a multi-channel guest in the maintainer Slack.
+
+### Emeritus Status
+
+When a former maintainer is offboarded, they will be automatically moved to Emeritus status within the governance repo. There is no minimum required period of servicing as a working group member for this status - any working group member who served for any length of time, but is no longer part of the working group, is eligible.
```
dsanders11, line 146:
```suggestion
When a former maintainer is offboarded, they will be automatically moved to Emeritus status within the governance repo. There is no minimum required period of serving as a working group member for this status - any working group member who served for any length of time, but is no longer part of the working group, is eligible.
```

dsanders11 (approved):

dsanders11:
I think we discussed maybe some language around it being a quicker process to return to full member status from emeritus? Not sure if we should add that in `charger/README.md`.

jkleinsc (approved):

MarshallOfSound (approved):

erickzhao (commented):
Sorry if we have discussed this elsewhere, but do we want to have emeritus members per WG or just for the project as a whole? Feel like we're gonna get a lot of duplication I guess between WGs.

dsanders11:
> Sorry if we have discussed this elsewhere, but do we want to have emeritus members per WG or just for the project as a whole? Feel like we're gonna get a lot of duplication I guess between WGs.

I kind of like the history that you get when they're per WG - learned a bit of WG history I didn't know. But, this brings up a good point regarding emeritus and defunct WGs - we can't capture those emeritus members since the WG directory no longer exists. So it is probably more complete to do emeritus overall rather than per-WG.

bnb (approved):

PR #578, opened 2024-07-24
chore(wg-ecosystem): add Node version policy and cleanup

erickzhao:
As discussed in today's meeting :)

### DIFF
--- a/wg-ecosystem/README.md
+++ b/wg-ecosystem/README.md
@@ -6,19 +6,19 @@ Oversees the projects that make Electron app development easier.
 
 | Avatar | Name | Role | Time Zone |
 | -------------------------------------------|----------------------|----------------------------| -------- |
+| <img src="https://github.com/blackhole1.png" width=100 alt="@blackhole1">  | Black-Hole [@blackhole1](https://github.com/blackhole1) | Member | BJT (Hangzhou) |
+| <img src="https://github.com/caoxiemeihao.png" width=100 alt="@caoxiemeihao">  | caoxiemeihao  [@caoxiemeihao](https://github.com/caoxiemeihao) | Member | BJT (Hangzhou) |
 | <img src="https://github.com/ckerr.png" width=100 alt="@ckerr">  | Charles Kerr [@ckerr](https://github.com/ckerr) | Member | CT (New Orleans) |
 | <img src="https://github.com/dsanders11.png" width=100 alt="@dsanders11">  | David Sanders [@dsanders11](https://github.com/dsanders11) | Member | PT (Santa Barbara) |
 | <img src="https://github.com/erickzhao.png" width=100 alt="@erickzhao">  | Erick Zhao [@erickzhao](https://github.com/erickzhao) | Member | PT (Vancouver) |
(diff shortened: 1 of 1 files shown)


wg-ecosystem/README.md
```diff
@@ -120,16 +115,28 @@ These repos are sorted alphabetically, their order does not reflect that any of
...
+## Node.js Upgrade Policy
+
+As a general rule, npm packages under the purview of the Ecosystem WG will support Node.js versions as
+long as possible, even after a version has reached its End-Of-Life (EOL) date.
+If too much friction arises from the burden of maintaining EOL versions of Node.js (e.g. trouble upgrading
+dependencies or CI issues), we will perform a blanket upgrade of supported Node.js versions for all our packages
+to the latest LTS version. This upgrade process will be revisited annually.
+
+As of July 2024, our current plan is to upgrade all packages to a minimum of Node.js 22 when that version
+reaches its LTS date ([see Node.js release schedule for more details](https://nodejs.org/en/about/previous-releases#release-schedule)).
```
dsanders11, line 130: I think we were also discussing jumping to 20.x if the ESM changes are backported. If we do 22 we'll probably want to wait N months after LTS rather than jumping to it right after it reaches LTS.

wg-ecosystem/README.md
```diff
@@ -6,19 +6,19 @@ Oversees the projects that make Electron app development easier.
 
 | Avatar | Name | Role | Time Zone |
 | -------------------------------------------|----------------------|----------------------------| -------- |
+| <img src="https://github.com/blackhole1.png" width=100 alt="@blackhole1">  | Black-Hole [@blackhole1](https://github.com/blackhole1) | Member | BJT (Hangzhou) |
```
MarshallOfSound, line 9: Why are we changing the order here? It was order-joined originally which was kinda nice

wg-ecosystem/README.md
```diff
@@ -60,10 +60,7 @@ These projects are sorted alphabetically, their order does not reflect that any
   * Update server (update.electronjs.org)
   * Windows Installer (`electron-winstaller`)
 * Tools (Userland)
-  * Devtron
-  * electron-compile
```
MarshallOfSound, line 64: Never 4get

wg-ecosystem/README.md
MarshallOfSound: Yeah maybe avoid specifying the version, say "the first LTS version with require-esm support" or something and link the node issue

wg-ecosystem/README.md
erickzhao: It was already alphabetized by GitHub username for everyone else but our most recent members added themselves at the bottom so I kind of fixed that.

dsanders11: That's true, considering I was second in the list. 😅

wg-ecosystem/README.md
erickzhao: [f9c421d](https://github.com/electron/governance/pull/578/commits/f9c421d6076f93a4bd06001cf27da91f2c612745)

wg-ecosystem/README.md
felixrieseberg: rip end of an era

felixrieseberg (approved):
Seems like what we did in fact discuss

PR #585, opened 2024-09-30
docs: type definition changes are not breaking

samuelmaddock:
Based on API WG discussions.

### DIFF
--- a/wg-api/best-practices.md
+++ b/wg-api/best-practices.md
@@ -12,6 +12,9 @@ API changes which alter existing behavior can cause apps to break unexpectedly w
 
 If the behavior must change to support the feature, the change should be listed in [the breaking changes document](https://github.com/electron/electron/blob/main/docs/breaking-changes.md). Additionally, consider whether the change can be introduced in a way which permits a deprecation cycle, for instance introducing the new API under a new name and deprecating the old name while keeping the behavior unchanged for apps using the API under the old name.
 
+> [!NOTE]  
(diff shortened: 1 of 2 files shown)


wg-api/best-practices.md
```diff
@@ -12,6 +12,9 @@ API changes which alter existing behavior can cause apps to break unexpectedly w
 
 If the behavior must change to support the feature, the change should be listed in [the breaking changes document](https://github.com/electron/electron/blob/main/docs/breaking-changes.md). Additionally, consider whether the change can be introduced in a way which permits a deprecation cycle, for instance introducing the new API under a new name and deprecating the old name while keeping the behavior unchanged for apps using the API under the old name.
 
+> [!NOTE]
+> Changes that only affect TypeScript type definitions are not considered breaking changes.
```
MarshallOfSound, line 16: This is tricky but I'd clarify it's not "only" rather "a change that breaks TS is not enough to be a breaking change, but still could be"

```suggestion
> Changes that break typescript apps by modifying TypeScript type definitions are not enough to consider a change breaking by itself. If the equivalent code written in JS would also break then it is considered breaking.
```

My words suck but I tried 😆

samuelmaddock: Updated to try to add clarity to runtime vs static compilation [25847cf](https://github.com/electron/governance/pull/585/commits/25847cf94aff2d303bbc204602a232b8aaa2237f)

MarshallOfSound (approved):

jkleinsc (approved):

PR #586, opened 2024-10-03
docs: define practices for third-party extensible APIs

samuelmaddock:
A few of Electron's existing APIs don't do a good job of supporting extensibility for third-party libraries. Let's try to define best practices to guide future API designs with third-party extensibility in mind.

### `protocol.registerSchemesAsPrivileged`

- Replaces all schemes without an option of appending
- Known issue for third-party libraries https://github.com/electron/electron/issues/40362
- Requires monkey patching to workaround ([see Sentry's example](https://github.com/electron/electron/issues/40362#issuecomment-2248468360))

### `session.setPreloads` / `session.getPreloads`

- Requires replacing all preloads rather than adding a single preload

```js
session.setPreloads([
  ...session.getPreloads(),
  path.join(__dirname, 'new-preload-script.js')
])
```

### DIFF
--- a/wg-api/best-practices.md
+++ b/wg-api/best-practices.md
@@ -156,11 +156,11 @@ protocol.registerSchemesAsPrivileged([
 ])
 
 // Good: third-party libraries can create, read, update, and delete items
-if (protocol.getRegisteredSchemes().includes(scheme => scheme.scheme === 'app')) {
-  protocol.unregisterScheme({ scheme: 'app' })
+if (protocol.getPrivilegedSchemes().includes(scheme => scheme.scheme === 'app')) {
+  protocol.unregisterSchemeAsPrivileged('app' )
 }
-protocol.registerScheme({ scheme: 'app', privileges: { standard: true } })
-protocol.updateScheme({ scheme: 'app', privileges: { secure: true } })
+protocol.registerSchemeAsPrivileged({ scheme: 'app', privileges: { standard: true } })
+protocol.updatePrivilegedScheme({ scheme: 'app', privileges: { secure: true } })
 ```
 
 ## Classes
--- a/wg-api/best-practices.md
+++ b/wg-api/best-practices.md
@@ -32,6 +32,13 @@ function whatever(opts: { a: string, b?: boolean }) { /* ... */ }
 
 See https://w3ctag.github.io/design-principles/#prefer-dict-to-bool for more details.
 
+### Should the configured options be made extensible to third-party libraries?
+
+If an API accepts configuring options, should it provide the ability to append to rather than replace those options?
+Would third-party libraries require special attention to API usage to avoid clobbering configured options?
+
+See the [style guide for designing APIs when dealing with arrays.](#provide-createreadupdatedelete-options-when-dealing-with-arrays)
+
 ### What underlying Chromium or OS features does this API rely on?
 
 If the API you’re changing relies on underlying features provided by Chromium or by the operating system, how stable are those underlying features? How might those underlying features change in the future?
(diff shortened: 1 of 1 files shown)


wg-api/best-practices.md
```diff
@@ -32,6 +32,13 @@ function whatever(opts: { a: string, b?: boolean }) { /* ... */ }
 
 See https://w3ctag.github.io/design-principles/#prefer-dict-to-bool for more details.
 
+### Should the configured options be made extensible to third-party libraries?
+
+If an API accepts configuring options, should it provide the ability to append to rather than replace those options?
+Would third-party libraries require special attention to API usage to avoid clobbering configured options?
```
itsananderson, line 38: I think this might actually apply to more than just configuring options, though that's the most obvious case. Here's some suggested tweaks to prompt us to think more broadly about how loosely coupled code will interact with an API. Feel free to edit further (or ignore) if you don't like my wording 😄 

```suggestion
### How will third-party libraries interact with this API?

When designing an API, consider that an app might have custom code and third-party libraries that both interact with that API. Can the API be designed so that multiple callers don't interfere with each other?

If an API accepts configuring options, should it provide the ability to append to rather than replace those options?
Can third-party libraries use the API without knowing what app-specific code is also using the API?
```

wg-api/best-practices.md
```diff
@@ -134,6 +141,26 @@ Even if seconds (or some other time unit) are more natural in the domain of an A
...
+protocol.registerSchemesAsPrivileged([
+  { scheme: 'app', privileges: { standard: true } }
+])
+
+// Good: third-party libraries can create, read, update, and delete items
+if (protocol.getRegisteredSchemes().includes(scheme => scheme.scheme === 'app')) {
+  protocol.unregisterScheme({ scheme: 'app' })
+}
+protocol.registerScheme({ scheme: 'app', privileges: { standard: true } })
+protocol.updateScheme({ scheme: 'app', privileges: { secure: true } })
```
itsananderson, line 161: I know this API doesn't actually exist, but if we were to implement this API, would there be a reason to have `unregisterScheme` take a dictionary rather than just being `protocol.unregisterScheme('app')`? I can't think of any other properties we'd ever want to add to this dictionary.

Also suggesting tweaks to API names since we're potentially going to point that PR at these docs as guidance for their API. There's an important distinction between registering schemes (can happen at any time) and registering schemes as privileged (must happen before `ready` event).

```suggestion
// Good: third-party libraries can create, read, update, and delete items
if (protocol.getPrivilegedSchemes().includes(scheme => scheme.scheme === 'app')) {
  protocol.unregisterSchemeAsPrivileged('app' )
}
protocol.registerSchemeAsPrivileged({ scheme: 'app', privileges: { standard: true } })
protocol.updatePrivilegedScheme({ scheme: 'app', privileges: { secure: true } })
```

wg-api/best-practices.md
samuelmaddock: Good point, and I like your phrasing better 👍

jkleinsc (approved):
LGTM

itsananderson (approved):
API LGTM

wg-api/best-practices.md
```diff
@@ -32,6 +32,15 @@ function whatever(opts: { a: string, b?: boolean }) { /* ... */ }
 
 See https://w3ctag.github.io/design-principles/#prefer-dict-to-bool for more details.
 
+### How will third-party libraries interact with this API?
```
alicelovescake, line 35: I wonder if the guidelines are applied to only 3rd party libraries? Perhaps we can make the title more generic like "How will different components interact with this API?" because I feel like this is helpful for the API author to think about not just 3rd party libraries, but also multiple internal modules interacting, or plugins or extensions.

timfish:
Thanks for your work on this!

PR #591, opened 2025-03-19
doc: add nikwen to Releases WG

clavin:
Adds @nikwen to the Releases WG.


MarshallOfSound (commented):
Can we please follow the onboarding workflow, this is like, step 3 I think

MarshallOfSound (approved):

PR #616, opened 2026-04-04
chore: resolve dependabot security alerts

MarshallOfSound:
Safe-only sweep of open Dependabot security alerts. Only changes that stay within existing semver ranges (or same-major resolutions) were applied — anything requiring a major bump of a direct dep is flagged below and left untouched.

### Resolved

| Package | Strategy | Version change |
|---|---|---|
| `glob` | scoped resolution (`glob@npm:~10.3.12` → `npm:10.5.0`) | `10.3.16` → `10.5.0` |
| `brace-expansion` | `yarn up -R` (in-range refresh) | `1.1.12` → `1.1.13`, `2.0.2` → `2.0.3` |

The `glob` alert comes in via `@electron/lint-roller@2.4.0` → `markdownlint-cli@^0.40.0` → `glob@~10.3.12`. There is no newer `2.x` of `@electron/lint-roller` and `markdownlint-cli`'s `~10.3.12` range can't reach the patched `10.5.0`, so a scoped resolution (same major) is the minimal safe fix. Only the `~10.3.12` descriptor is overridden — the `7.x`/`8.x`/`9.x` copies of `glob` elsewhere in the tree are untouched.

### Flagged (not changed)

- **`@electron/lint-roller` `^2.1.0` → `^3.x`** — would remove the `markdownlint-cli@0.40.0` chain entirely (and pull `glob@^10.4.5` directly), but it's a major bump of the only direct dep so out of scope for a safe-only sweep. Worth doing as a follow-up, which would also let the `glob` resolution be dropped.
- **`markdown-it` (audit only, not an open Dependabot alert)** — `13.0.2` / `14.1.0` in tree, patched in `14.1.1`. The `13.x` copy is pinned by `@electron/lint-roller@2.4.0` (`^13.0.1`) and the `14.1.0` copy is exact-pinned by `markdownlint@0.34.0`; both clear with the `lint-roller` `3.x` bump above.

### Verification

- `yarn install --immutable` passes
- `yarn npm audit`: `glob` and `brace-expansion` advisories cleared; only `markdown-it` (moderate, not an open Dependabot alert) remains

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,5 @@
     "@electron/lint-roller": "^3.3.0",
     "markdownlint-cli2": "^0.22.0"
   },
-  "resolutions": {
-    "glob@npm:~10.3.12": "npm:10.5.0"
-  },
   "packageManager": "yarn@4.10.3+sha512.c38cafb5c7bb273f3926d04e55e1d8c9dfa7d9c3ea1f36a4868fa028b9e5f72298f0b7f401ad5eb921749eb012eb1c3bb74bf7503df3ee43fd600d14a018266f"
 }
--- a/package.json
+++ b/package.json
@@ -5,5 +5,8 @@
     "@electron/lint-roller": "^3.3.0",
     "markdownlint-cli2": "^0.22.0"
   },
+  "resolutions": {
+    "glob@npm:~10.3.12": "npm:10.5.0"
+  },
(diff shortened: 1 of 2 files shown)


dsanders11:
I think we should land #615 first.

VerteDinde (approved):

VerteDinde:
Approving since we landed https://github.com/electron/governance/pull/615

PR #621, opened 2026-06-23
docs: clarify event object usage in EventEmitter examples

samuelmaddock:
I noticed we don't detail the pattern of including details on event objects. This was brought up in an API review for https://github.com/electron/electron/pull/51563#discussion_r3455401277


itsananderson (commented):
Is it worth also describing (or linking to an example that demonstrates) how to *document* an API to accept an `Event` with extra properties? Unlike other examples in this Best Practices doc, it may not be obvious how to write the API docs so that the correct type interface is generated.

PR #634, opened 2026-09-20
build: update @electron/lint-roller to 4.0.0

MarshallOfSound:
Bumps `@electron/lint-roller` to 4.0.0, which drops the transitive eslint/standard dependency tree (yarn.lock shrinks accordingly). The shared markdownlint config and custom rules are unchanged.

Two knock-on changes in the lint workflow: lint-roller 4 needs Node 22.17 or newer, so the job moves from Node 20 to 24; and `lint-roller-markdown-links` no longer has a `--fetch-external-links` mode (electron/lint-roller#203, too many sites block automated fetches), so the manual "Check external links" dispatch input and step are removed. Relative link and fragment checking is unaffected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GaskPz9P37WZvpQ5zKit7V)_


MarshallOfSound:
@SocketSecurity ignore npm/@electron/lint-roller@4.0.0

